PR #183, opened 2026-05-02
fix(ops): bind-mount pipeline + ros2/src so host edits go live (#176)

### LINKED ISSUE #176: ops: rebuild dv_pipeline_stack image after every PR that touches pipeline/* — docker compose silently reverts to image source on restart
## Problem
\`docker compose up -d\` (re-)creates the dv_pipeline_stack container from the image's baked-in source tree. The image was built at some past point and **does not contain any of the rewrites since then**:
- new control framework (state.py, reference.py, controllers/, models/) — feat/34, feat/55, fix/58, fix/60, fix/61
- Delaunay planner — feat/35
- Lichtblick layout topic additions — feat/35

If the image is stale, the container reverts to the OLD code on restart. Symptom we hit today: \"car won't drive after Start Session\" (Plan_Path returned empty paths because it was the pre-Delaunay forward-walker built against a different cone field that didn't match the new test_submodule layout).

## Workaround used during development
\`docker cp <host file> ifssim-dv_pipeline_stack-1:<container file>\` for every changed Python file, then \`colcon build --symlink-install --packages-select <pkg>\` inside the container, then restart the pipeline. Survives until the next \`docker compose down\`.

## Fix
After merging any PR that touches:
- \`pipeline/**\` (slam, cone_slam, path_planning, control)
- \`docker/dv_pipeline_stack/**\` (entrypoint, launch files)
- \`ros2/src/**\` (bridge / fs_msgs)

Rebuild the image:

\`\`\`
docker compose build dv_pipeline_stack
docker compose up -d --force-recreate dv_pipeline_stack
\`\`\`

Or as a one-liner:

\`\`\`
docker compose up -d --build dv_pipeline_stack
\`\`\`

## Better long-term (separate work)
Bind-mount \`pipeline/\` and \`ros2/src/\` into the container so source edits go live without rebuilding. The current setup mounts only \`pipeline_ctrl\` and \`numba_cache\` as named volumes; \`COPY pipeline/ src/\` in the Dockerfile means the source is frozen at image build time.

A bind-mount approach trades a bit of build speed (colcon must run inside the container at startup since \`--symlink-install\` is already used) for restart-safety: any local change is automatically present in the next container start.

This isn't blocking, just slow. File a follow-up if anyone wants to pick it up.

raulmoranguerra:
## Summary

Closes #176. The dv_pipeline_stack container baked pipeline/ and ros2/src/ source into the image at build time, so any restart silently reverted to whatever code was current at the last image build. Today's "empty path" episode was the latest instance — the running container was still on the pre-Delaunay forward-walker even though `dev` had merged the new planner three commits prior.

The image was already built with \`colcon build --symlink-install\`, which makes the install/ tree's Python entries symlinks back to \`/dv_pipeline_stack_ws/src/<pkg>\`. We just weren't keeping that source live at runtime. Now we are: each ROS package is bind-mounted from the host over the image's COPY'd baseline.

## Changes

- **docker-compose.yml**: 6 new bind-mounts on the dv_pipeline_stack service — one per package (`./pipeline/cone_slam`, `./pipeline/control`, `./pipeline/path_planning`, `./pipeline/slam`, `./ros2/src/fs_msgs`, `./ros2/src/ifssim_bridge`). Mounted per-package because \`pipeline/\` and \`ros2/src/\` collapse into a single \`src/\` inside the container — a single parent-level mount of either would hide the other. Plus a new \`DV_REBUILD_ON_STARTUP\` env passthrough.
- **entrypoint.sh**: opt-in \`DV_REBUILD_ON_STARTUP=true\` triggers \`colcon build --symlink-install\` before the install/ source. Off by default. Use it when you've changed setup.py, added a package, edited a .msg, or touched C++.
- **docs/GETTING_STARTED_DOCKER.md**: rewrite the "Rebuilding after code changes" section to make explicit (a) Python edits are zero-rebuild, (b) C++/msg/setup.py changes still need a rebuild, (c) two ways to trigger one.

## Workflow after this

| Change | What you do |
|---|---|
| Python edit in existing pipeline file | Toggle the pipeline (\`/api/event/start\` or \`rm /pipeline_ctrl/enable && touch /pipeline_ctrl/enable\`). Done. |
| New file / new entry_point in setup.py | \`docker compose exec dv_pipeline_stack colcon build --symlink-install --packages-select <pkg>\`, then toggle pipeline. Or \`DV_REBUILD_ON_STARTUP=true docker compose up -d --force-recreate dv_pipeline_stack\`. |
| C++ change in ifssim_bridge | Same as above. |
| .msg change in fs_msgs | Same as above. |

## Test plan

- [x] \`yaml.safe_load\` validates the compose file; all 6 mounts present.
- [x] \`bash -n\` validates the entrypoint syntax.
- [ ] Live test: stop containers, \`docker compose up -d --force-recreate dv_pipeline_stack\`, \`docker compose exec dv_pipeline_stack ls -la /dv_pipeline_stack_ws/src/cone_slam\` should show host paths via bind-mount; touch a .py file on the host, restart pipeline, verify the change is in effect (e.g. add a log line and watch for it).
- [ ] Verify \`DV_REBUILD_ON_STARTUP=true\` runs colcon at startup and sets up correctly afterwards.


raulmoranguerra:
Live test passed on Docker (Apple Silicon, just now):

- All 6 bind-mounts visible inside the container (`fakeowner` mounts under `/dv_pipeline_stack_ws/src/{cone_slam,control,path_planning,slam,fs_msgs,ifssim_bridge}`).
- Host edit propagation: flipped `SEARCH_RADIUS_M = 7.0 → 99.9` on host with `sed`; `python3 -c 'from path_planning.planner import SEARCH_RADIUS_M'` inside the container immediately printed `99.9`. Reverted on host, container re-imports `7.0`. No rebuild, no `docker cp`, no container restart.

The install/ → src/<pkg> symlinks from `--symlink-install` now resolve through the bind-mount as designed. Ready to merge.

Issue #157, opened 2026-05-01
Cones fall through the level floor at spawn


raulmoranguerra:
Cones (yellow / blue / big-orange spline-spawned cones) sometimes fall through the level floor at spawn time, observed during baseline-tuning runs on \`ci_test.csv\`.

Likely causes:
- Cone static mesh / Blueprint has a rigidbody but the level static mesh under it has \`Collision Enabled = QueryOnly\` instead of \`QueryAndPhysics\`
- Cones spawn at z=0 with their pivot below the floor
- Track import didn't elevate cones to floor height

UE-side fix only (asset / level setup). No code change from the autonomy stack needed.

## What to verify in the editor
- Floor's static mesh \`CollisionEnabled\` should be \`QueryAndPhysics\` (not \`QueryOnly\`)
- Cone Blueprint should spawn at z = floor_height + cone_pivot_offset, not z=0
- Possibly \`bSimulatePhysics\` on cones is wrong for spawn-time stability — disable physics, enable on contact only

Track-load logs report \`car_aligned:false\` for some tracks — might be related.

raulmoranguerra:
Largely fixed via PR #185 (merged in commit 372988b). The CSV cone-spawn path now line-traces down at each cone's XY, places the mesh-base on the actual floor surface plus a HeightOffset clearance, and emits a per-track-load summary so future regressions are visible.

Verified on customMap with test_submodule.csv (168 cones): `FSDS ConeSpawner: ground-snap 168 hit / 0 missed across 168 cones` — all cones sit on the floor, none sunk. The previously-observed far-row sink is gone.

What's not covered (separate from this issue):
- Spline-spawned cones (BP_spline_cones_*) — different code path, not touched by PR #185.
- A floor static mesh with `Collision Enabled = QueryOnly` could still let physics drag cones through it after gravity engages, even with correct spawn placement; that's an editor-side asset fix on whichever map's floor is misconfigured.

Reopen if either case bites in practice.

PR #185, opened 2026-05-02
fix(spawn): ground-snap CSV-spawned cones via line trace (partial #157)

raulmoranguerra:
## Summary

Partial fix for #157. CSV-spawned cones were planted at a flat \`HeightOffset\` (default 5 cm) regardless of the actual floor under each XY, then the Referee enabled physics immediately after spawn. When the cone mesh's bottom poked into the floor at t=0, Chaos's penetration-resolution had to eject it — and on a floor whose collision is \`QueryOnly\` rather than \`QueryAndPhysics\`, the eject had nothing to push against and the cone phased through.

This change stops spawning cones below the floor: a downward line trace at each cone's XY finds the actual surface, and the cone is placed so its mesh's lowest point sits on the floor + \`HeightOffset\` clearance.

## What's covered
- CSV-spawned cones (the path the autonomy uses for every loaded track) now ground-snap via \`LineTraceSingleByChannel\` against \`ECC_WorldStatic\`.
- Uses \`Mesh->GetBounds()\` in local space + \`ConeScale\` so it works regardless of where the cone mesh's pivot sits.
- Trace miss (level edge, hole, no floor below) falls back to the original \`HeightOffset\`-only behaviour and logs a warning so it's visible if it fires.

## What's not covered (editor-side, can't fix from C++)
- **Floor \`Collision Enabled = QueryOnly\` instead of \`QueryAndPhysics\`.** If the floor's collision profile is wrong, runtime physics will still pull cones through it once gravity engages — the ground-snap above only guarantees they aren't *born* below the floor. Needs an editor-side asset fix on the level's floor static mesh.
- **Spline-spawned cones** (\`BP_spline_cones_*\`). Those go through a Blueprint path, not \`SpawnStaticMeshCone\`, and need their own audit in the editor.

#157 stays open until those editor-side checks are done.

## Build verification
\`\`\`
$ Engine/Build/BatchFiles/Mac/Build.sh IFSSIMEditor Mac Development -Project=...
[1/3] Compile [Apple] FSDSConeSpawner.cpp
[2/3] Compile [Apple] Module.FSDSPlugin.cpp
[3/3] Link [Apple] UnrealEditor-FSDSPlugin.dylib
Result: Succeeded
\`\`\`

## Test plan
- [x] UBT build of \`IFSSIMEditor\` succeeds (above).
- [ ] Live PIE: load a track, watch for \`FSDS ConeSpawner: Spawned N cones\` log; visually confirm no cones below the floor and that the \`ground line-trace missed\` warning does *not* fire.
- [ ] If any cones still go through the floor, that's the editor-side floor-collision problem documented above — file as a follow-up issue with the affected map.


raulmoranguerra:
Live PIE verification on customMap (fresh editor restart, new dylib loaded):

```
FSDS ConeSpawner: ground-snap 168 hit / 0 missed across 168 cones
```

All 168 cones snapped to the floor via line trace — zero misses, zero warnings. User confirmed cones sit on top of the floor in the viewport (previously the far row was sunk into the ground; now they're at proper height). Merging.

Issue #151, opened 2026-05-01
Control: steering dynamics — servo rate-limit + first-order lag


raulmoranguerra:
Currently \`SetSteeringInput\` is applied instantly each tick. The autonomy can issue rate-unlimited commands the sim obeys without delay, masking real EPAS dynamics that the IFS-08 will exhibit on track.

## What to do
Model the steering actuator as:
- **Slew rate limit** ~5–10 rad/s at the column (TBD from EPAS spec)
- **First-order lag** ~30 ms time constant (TBD)

Apply to \`CurrentControls.Steering\` before passing to \`VehicleMovement->SetSteeringInput\` in \`AFSDSVehiclePawn::Tick\` ([FSDSVehiclePawn.cpp](Plugins/FSDSPlugin/Source/FSDSPlugin/Private/FSDSVehiclePawn.cpp)).

## Acceptance
- Step input from 0 → 1 reaches >95% of target in ≥ 4×τ ≈ 120 ms (not one tick)
- Slew limit prevents servo from exceeding spec rate when commanded a step
- Real-IFS-08 EPAS spec values referenced in code comments

Spawned from #148.

raulmoranguerra:
Closing — **partially landed in #308, remainder moves to the submodule's `control_node`** (see #311 freeze and `docs/integrated_architecture.md`).

Original scope: servo rate-limit + first-order lag on the steering channel.

**Rate-limit half (the substantive bug):** landed in #308. `control_node` now applies a slew-rate limit at the publish boundary (`steering_rate=5.0` /s default, ≈ 600 °/s on a real FS rack; same mechanism applied to throttle and regen). Live-traced before/after on `test_submodule`:

- Before: steering swung to ±1.0 in adjacent ticks during the GT-as-SLAM diagnostic run (#306).
- After: the same path produces gradient-bounded steering with no thrash, matching real-car servo dynamics.

The math is in `control_node._tick` and the parameter constants in `_declare_params`; the helper is the bottom-of-file `_slew(prev, target, max_step)`. It's strategy-agnostic so it ports to any future controller.

**First-order lag half:** the rate-limit alone proved sufficient on the lap traces we have, so the lag was deferred. On the new architecture the entire `control_node` belongs to the submodule. If the submodule team wants to add lag on top of the slew-limit, they have two clean options:

1. Apply a discrete first-order filter to each channel inside the same publish-boundary block where `_slew` lives. Cost: 3 floats of state, 2 multiplications per tick.
[...]

Issue #150, opened 2026-05-01
Control: traction control — slip-ratio cap on EMRAX torque command


raulmoranguerra:
At thr=1.0 launch we observe slipMag ~270 cm/s (~2.5 m/s wheelspin) on the rear axle. EMRAX delivers full envelope torque regardless — the friction circle clips it to AvailableGrip but the leftover torque just spins the wheel uselessly. A real-car ECU would close the loop on slip ratio.

## What to do
Add a slip-ratio cap to the EMRAX command in [FSDSVehiclePawn.cpp Tick](Plugins/FSDSPlugin/Source/FSDSPlugin/Private/FSDSVehiclePawn.cpp):
- Compute per-wheel slip ratio σ = (ω·r − v_chassis) / max(v_chassis, ε) using wheel ω from \`VehicleMovement->Wheels[i]->GetWheelAngularVelocity()\`
- Target peak σ ≈ 0.10 (Pacejka peak μ_x for Hoosier R20)
- When σ > σ_target, scale down throttle command going into \`Motor->Step()\` proportionally

This is a separate knob from Chaos's built-in \`bTractionControlEnabled\` (which we leave off — it's an idealised cap, not a real ECU loop).

## Acceptance
- thr=1.0 launch shows slipMag converging toward ~50 cm/s (peak slip), not 270+
- Steady-state ax stays >= post-fix/54 level (3.9 m/s²) — TC reduces wasted torque, doesn't remove available grip
- Comments link to Pacejka peak σ rationale

Spawned from #148.

raulmoranguerra:
Closing on the IFSSIM side — moves to the submodule's `control_node` (see #311 freeze and `docs/integrated_architecture.md`).

Why moved, not just closed:

- This issue asked for a slip-ratio cap on the EMRAX torque command inside our PI velocity controller. That controller is being re-authored in the submodule and the cap belongs in the new `control_node`, not in the frozen in-tree one.
- The new architecture's strategy pattern on lifecycle nodes (the team's "patrón strategy para variar comportamiento entre las distintas pruebas dinámicas") is the natural home for traction control — it's mission-specific (more aggressive on accel, more conservative on skidpad's tight radius) and per-mission strategy is exactly what `mode_manager_node` already passes via the mission flag.

Notes worth carrying forward for the submodule team:

[...]

PR #187, opened 2026-05-02
fix(planner): drop same-color Delaunay edges to fix wrong-side selection in tight corners (#180)

### LINKED ISSUE #180: planner: Delaunay path breaks on tight turns — needs tuning
## Symptom
Observed during the trackdrive test that closed #177: the Delaunay-based planner produces a clean centerline on straights and gentle curves, but on **tight turns** it breaks — the path becomes empty, jumps, or selects a midpoint that puts the car on the wrong side of the corridor.

## What we know
- The Delaunay + best-first-search MVP (PR #167) was tuned for FS-standard 5 m start-gate spacing and ~3 m straight-track cone spacing.
- During the verified end-to-end run, steering hit ±1.0 (saturated) on the tightest sections — that's likely the same regions where the planner was either dropping out or feeding the controller a bad reference.
- Current planner constants in [planner.py](pipeline/path_planning/path_planning/planner.py):
  - `SEARCH_RADIUS_M = 7.0` — candidate gather radius around each leaf
  - `MAX_HEADING_DELTA_RAD = 0.8` (~46°) — per-step heading change cap
  - `MIN_TRACK_WIDTH_M = 2.0` — candidate's edge must be ≥ this long
  - `MIN_MIDPOINT_DIST_M = 0.81` — min step from leaf to next midpoint
  - `MAX_EDGE_LEN_M = 7.0` — drop triangles with any edge longer than this
  - `MIN_TRI_ANGLE_RAD = 0.35` (~20°) — drop sliver triangles
- FS rules give a minimum corner radius around 4.5 m (TI 3.1.x); cone spacing on tight corners can drop to ~2.5 m on the inside line. Several of these constants were chosen against straight-track fixtures and may not survive the geometry of a tight turn.

## Hypotheses to investigate
1. **`MAX_HEADING_DELTA_RAD = 0.8` clips legitimate tight-corner steps.** A 4.5 m radius corner with 2.5 m cone spacing demands ~32°/step on the inside line — close to the cap and probably exceeded once dropouts compound. Symptom: planner refuses the only valid next candidate and returns empty.
2. **`MIN_TRACK_WIDTH_M = 2.0` rejects valid blue/yellow pairs on the apex** where the inside cone is closer to the spline. FS minimum track width is 3.0 m, but the *Delaunay edge length between left and right cones* on a tight corner can be shorter than 3 m because the two cones span the chord, not the centerline.
3. **`MIN_TRI_ANGLE_RAD = 0.35` filter drops the apex triangle.** On a tight turn the apex triangle's interior angle at the apex cone is small by construction — the geometry is genuinely a sliver. Need to either relax the angle filter on triangles whose pair-edge is short, or use a different sliver test (e.g. aspect ratio).
4. **Best-first walker has no replanning.** When the walker hits a dead end inside a turn, it terminates rather than backtracking. Adding 1-step backtrack would make tight corners robust to a single bad candidate selection.
5. **PCHIP densification artefacts on short midpoint chains.** With only 3–4 midpoints inside a tight turn, the spline is under-determined and can overshoot.

## How to investigate
1. Capture `/path_planning/delaunay` (the debug overlay PR #167 added) on a corner where the path breaks. Check whether the triangulation contains the apex triangle, whether candidate midpoints exist, and whether the walker selected them or filtered them out.
2. Add per-step rejection-reason counts to `plan_centerline_with_debug` (heading-delta, track-width, distance-cap) so we can localise which gate fires.
3. Build a `test_tight_turn.py` fixture in [test_planner.py](pipeline/path_planning/test/test_planner.py): R = 4.5 m, cone spacing 2.5 m on inside / 3.5 m on outside. Assert the path completes.

## Acceptance
- Planner produces a complete centerline through a synthetic 4.5 m radius corner without dropping out.
- Verified in PIE on the existing test track's tightest corner — no empty-path windows ≥1 s during a full lap.

raulmoranguerra:
Closes #180.

## Root cause

On a tight FS-minimum-radius right turn under realistic LiDAR position noise (~10–20 cm σ), the planner sometimes picked a path that went the *wrong way* through the corner — end yaw +7.6° on a corner whose centerline reaches −85°. Reproducible with a synthetic 90° R=4.5 m corner at noise σ=0.15 m seed=0.

The selected midpoint was at (8.27, +0.51) — produced by a Delaunay edge between two BLUE cones (one before the apex, one past it) spanning across the corner. That midpoint sits roughly along the straight extension of the approach. The greedy heading-biased score (W_DIST=0.1, dhead-term dominates 9:1) prefers it overwhelmingly: dhead ≈ 2° vs ≈ 30–45° for any real corner midpoint. Once the chain is steered along that phantom edge, the heading-delta cap keeps it locked into the wrong-direction trajectory.

Same-color edges aren't centerline crossings by construction: the centerline is *between* blue (left) and yellow (right); an edge between two same-color cones is along one side, and its midpoint sits on that side, not in the corridor.

## Fix

\`_build_edges\` now takes per-cone colors and skips edges whose endpoints share a color. Triangulation itself is unchanged — only the candidate-midpoint set is filtered before the search.

This also cleans up the rejection-counter output: the previously-massive \`track_width\` bucket (which was mostly catching short same-side same-color edges through the straight approach) shrinks from ~46 to 0 on the standard fixture, because those edges are filtered earlier.

## Verification — synthetic 90° R=4.5 m corner, end yaw across noise

| Case               | Pre-fix      | Post-fix     |
|--------------------|--------------|--------------|
| Noiseless          | −31.1°       | **−84.6°**   |
| σ=0.10 m seed=2    | **+25.6°** ✗ | **−83.0°**   |
| σ=0.15 m seed=0    |  **+7.6°** ✗ | **−84.5°**   |
| σ=0.20 m seed=0    |  **+4.6°** ✗ | **−84.6°**   |

(Pre-fix +ve end yaw on a right turn means the path inverted through the corner.)

## Diagnostic instrumentation also added

\`PlanDebug\` gains a \`rejections: dict\` field that \`_search_path_body\` populates with per-gate counts (\`distance\`, \`track_width\`, \`heading_delta\`, \`behind_leaf\`, \`self_intersect\`). Useful when a path comes back too short or empty — points straight at which gate to relax. The new tests assert against this dict in their failure messages so a future regression prints the bottleneck gate without re-running with a debugger.

## Tests

- \`test_tight_corner_45_deg\` — moderate right corner; sanity case.
- \`test_tight_corner_90_deg\` — full FS-minimum corner. Pre-fix reached only −31°; assertion now requires < −60° (path actually reaches −84.6°).
- \`test_tight_corner_with_noise_doesnt_invert\` — 4 noise levels × 4 seeds = 16 cases, none may invert (end yaw must stay < −5° on a right turn). Three of these inverted pre-fix.

10/10 tests pass.

## Out of scope (worth a follow-up tuning pass)

The probe also showed two secondary failure modes that this PR doesn't fix:
- Some seeds (noise≥0.05 m seed=1) stop at end yaw ≈ −33° instead of full −84°. Path completes but stops short.
- Mid-corner observation at body yaw ≈ −60° gives empty path (only 2 selected midpoints, < MIN_MIDPOINTS=3).

Both look like score-function tuning rather than algorithm flaws. Worth revisiting after we've driven a real lap with this fix in place.

## Test plan

- [x] All 10 unit tests pass (\`docker compose exec dv_pipeline_stack pytest\`).
- [ ] Live PIE: drive a full lap on the existing test track with the fix; verify no empty-path windows ≥ 1 s during corner traversal (issue acceptance criterion).


raulmoranguerra:
## Update — actual root cause found via captured PIE data

The earlier same-color-edge filter on this PR fixed the synthetic test but **did not solve the real PIE failure**. Captured a 153-tick PIE run, traced the actual root cause, and pushed two more commits:

1. **First-step heading constraint** — the walker's first step had `dhead = 0` (no previous heading), so any cross-color midpoint near the car won on distance alone. Now the first step compares heading to body-+x (the car's forward direction). An off-centerline midpoint at body-y=−1.19 has dhead ≈ 41°, an on-centerline one at +0.43 has dhead ≈ 8° — second wins.

2. **MAX_HEADING_DELTA_RAD relaxed from 0.8 → 1.1 rad** (~46° → ~63°). On consecutive sharp curves where SLAM has caught one side of the corridor but not the other, the previous cap dead-ended the search.

Also added: **DV_PLANNER_CAPTURE** env var (path_planning node dumps one JSON line per callback to a file) + per-gate rejection counters in `PlanDebug.rejections`. Together these let me localise the exact failure within minutes instead of bisecting through PIE.

### Replay results on captured PIE data

| Capture (#ticks)            | plan_empty before/after | sel<4 before/after |
|-----------------------------|-------------------------|---------------------|
| Original failure (153)      | 14 (9.2%) → **0**       | 20 (13.1%) → **1**  |
| After first-step fix (196)  | 0 → **0**               | 25 (12.8%) → **0**  |

### Live PIE re-test

Car drove cleanly through the first turn — `plan_empty=0` across 159 ticks, steering moderate (−0.5 to −0.2), never saturated. The car's eventual stop at +42 m is downstream of the planner (with 20 m of healthy reference path) and looks like a controller/physics issue (#150 / #151) rather than a planner one.

### Tests

Added `test_pie_capture_tick_recovers` — replays the actual failing tick 84 (39 cones, colour-confused layout). Pre-fix: empty path. Post-fix: 10 selected midpoints.

11/11 tests pass. Ready for re-review.

raulmoranguerra:
## Final scope of this PR

Live PIE testing today found that even with this PR's fixes the car still crashes around the corner exit, but the failure mode is **upstream of the planner**: the SLAM colour classifier locks cone colour on first observation, and in tight corners that locks many cones to the wrong colour (filed as #188 with hypothesis + four candidate fixes). With the colour labels broken, no amount of planner work fixes the path going through the cone row.

### What this PR actually accomplishes (verified)

- **First-step heading constraint** stops the walker picking a colour-confused near midpoint and dead-ending one step later. Captured tick 84 from a real PIE failure: pre-fix `sel=2` → `path=0` (empty), post-fix `sel=10` → `path=30`.
- **MAX_HEADING_DELTA_RAD relaxed 0.8 → 1.1 rad**: on the second corner, the previous cap dead-ended the walker. Post-fix, full path through the second corner.
- **Same-colour penalty (soft)**: keeps cross-colour midpoints preferred when both are available. No hard drops.
- **PlanDebug.rejections diagnostic dict**: per-gate rejection counts. When a path comes back too short, the bottleneck gate is one log line away.
- **DV_PLANNER_CAPTURE env var**: per-tick JSONL dump of (cones, pose, n_path, rejections). Used to bisect this exact failure offline; will be the standard way to debug planner regressions going forward.

### Replay-on-real-data results

| Capture (#ticks) | plan_empty before/after | sel<4 before/after |
|---|---|---|
| Original (153 ticks at v_max=3) | 14 (9.2%) → **0** | 20 (13.1%) → **1** |
| After first-step fix (196) | 0 → **0** | 25 (12.8%) → **0** |
| Latest at v_max=1.5 (316) | 0 → **0** | 0 → **0** |

### What this PR does NOT fix
- Car still crashes into cones around the corner exit. Tracked as #188 — needs a SLAM-side fix, not a planner fix.

### Recommendation
Merge this for the planner improvements (real, captured-data-verified, regression-tested) and leave #188 as the focused next investigation.

PR #192, opened 2026-05-02
fix(planner): trim same-colour tail so path doesn't extend through one-sided cone region (#189)

### LINKED ISSUE #189: planner: path runs through cone row when only one side of corridor is observed
## Symptom

After #180/#187 the planner produces stable, full-length paths (0 plan_empty across 316 ticks). But on a tight left turn on `test_submodule.csv` the path runs along the BLUE cone row instead of between blue and yellow, so the car physically clips cones at the corner exit (~50 m from spawn).

Lichtblick screenshot from the failing session shows blue cylinders forming a continuous arc, with the planner's path running coincident with that arc.

## Root cause (confirmed via DV_SLAM_LANDMARK_CAPTURE — see #188)

It's NOT a colour-classifier bug. Every landmark the SLAM creates has its body-y unambiguously on the correct side of the car at first observation — closest near-boundary case had body_y = −0.51 (clearly YELLOW). The classifier is doing the right thing.

**The actual cause is geometric observation asymmetry.** At the crash point, the SLAM map contains:
- BLUE: 28 landmarks spanning world-x 5–47 m
- YELLOW: 8 landmarks, all at world-x ≤ 22 m

The LiDAR (Hesai ATX, ±60° H FoV, ±18° V FoV, range ~25 m) sees the outside arc of the turn (which is BLUE for this left turn) at long range. The inside arc (YELLOW) is closer and has fewer cones in any single FoV snapshot — and as the car drives along, inside-arc cones go past the LiDAR's rear before equally-many new ones come into the front view.

End result: in world-x ∈ [22, 47] there's a region of the SLAM map with only BLUE cones. Delaunay over that region produces only same-colour edges. The planner's same-colour penalty (W_SAME_COLOR_PENALTY = 1.5 from #187) is enough to prefer cross-colour midpoints when both kinds are available, but here cross-colour edges don't exist past world-x = 22. The walker picks blue-blue midpoints, which sit on the blue arc.

## Possible fixes (rough — needs the real investigation)

1. **Bound the path's lookahead to the cross-colour region.** Find the last edge with one BLUE and one YELLOW endpoint along the walker's chain; truncate the spline output there. Means shorter paths during one-sided observation; controller has to handle short paths. Cleanest semantically.

2. **Project a virtual centerline forward when one side runs out.** Take the spacing/curvature of the last few cross-colour midpoints and extrapolate. Risky on curves.

3. **Drive slower into the corner so more cones get observed before the LiDAR moves past.** Tune `a_lat_max` / `v_max` in the longitudinal controller to brake more aggressively into the turn. Doesn't fix the planner but may sidestep the failure in practice.

4. **Use the second LiDAR / a wider FoV.** Sim has only one Lidar1 at the moment. A rear-or-side LiDAR would capture inside-arc cones longer.

5. **Persist the observed cross-colour region.** Even after a yellow cone leaves the FoV behind the car, its position is still in the SLAM map. The planner already uses the full map (not just visible cones) — but past the corner the inside cones literally aren't behind the car yet, they haven't been observed at all.

## Diagnostic infrastructure already in place
- `DV_PLANNER_CAPTURE` (from #187) — per-tick (cones, pose, rejections).
- `DV_SLAM_LANDMARK_CAPTURE` (from #188 investigation, will land in the next PR) — per-landmark (body_x, body_y, range, bearing, colour, pose_yaw) at creation.

Both written as JSONL to a host-bind-mounted `/pipeline_ctrl/*.jsonl`, replayable offline.

## Acceptance
- A full lap on `test_submodule.csv` without the car physically clipping cones.
- The planner's path stays inside the corridor (between blue and yellow) for every tick of the lap, not just where both sides are observed.

raulmoranguerra:
## Summary

Closes #189. Trims the trailing run of same-colour midpoint picks from the walker's chain before returning. Same-colour midpoints sit on a single cone row, not between two rows; allowing them at the path's tail caused the spline to project *through* the BLUE cone arc on tight corners with one-sided LiDAR observation, and the car physically clipped cones.

## How it works

The walker still picks same-colour edges as fallback (with the soft penalty from #187). After the chain is built, a back-walk finds the last cross-colour pick and truncates everything past it.

```
c c c s s s   →  c c c
c s c c s     →  c s c c
s s s s s     →  no trim possible
```

**Soft trim**: if trimming would leave fewer than MIN_MIDPOINTS picks (3), the original chain is kept. A degraded path is better than empty — the controller has *something*, and the next tick (with fresh observations) usually recovers.

### Replay on captured PIE data (the v3 trace that originally crashed)

| Metric | Before | After |
|---|---|---|
| plan_empty ticks | 0 | 0 |
| sel<4 ticks | 0 | 0 |
| sel distribution | (5,15),(6,33),...(11,20),(12,1) | (4,16),(5,16),(6,64),(7,56),(8,7) |

Right tail of long chains gone (those were the same-colour projections); cross-colour core intact.

### Live PIE on test_submodule.csv at v_max=3.0

```
t=00s  pos=(  0.00, +9.01)  spawn
t=14s  pos=( +3.85, -19.28)  exiting straight
t=22s  pos=(+19.56, -35.94)  through first corner
t=30s  pos=(+33.90, -24.95)  past second section
t=42s  pos=(+58.56,  -9.93)  far end
t=50s  pos=(+41.00,  +8.43)  approaching start back
t=52s  pos=(+14.76, +35.38)  *** SLAM jump (out of scope) ***
```

**~200 m of autonomous trackdrive across multiple corners.** Approximately a full lap before a SLAM divergence event ended the run. Every previous test stopped at the first or second corner ~30-40 m from spawn.

## Out of scope

- The SLAM divergence at t≈52 s (sudden +25 m position jump). Looks like an iSAM2 cascade — separate investigation.
- Controller fidelity at v=3 m/s (#150, #151) — margin is tight on tight corners; the car can follow the path but a slip-ratio cap and steering rate-limit would help.

## Tests

- 11/11 pass.
- test_pie_capture_tick_recovers: previously 10 picks, now 3 (seed + 2 cross-colour). Path stays inside the corridor.


raulmoranguerra:
Closing without merging. Both attempted variants (soft and hard same-colour-tail truncation) had unacceptable side effects:

- **Hard trim** (drop tail unconditionally): 58% of replay ticks went empty; in PIE the car never moved at all, even from spawn (the walker often picks same-colour edges early when the field is small/asymmetric).
- **Soft trim** (keep chain when trim leaves < MIN_MIDPOINTS): the car drove ~200 m vs the previous ~30 m, but on tight one-sided sections the path still extended through the cone arc and the car still hit cones — same root symptom #189 reports.

The diagnostic infrastructure from this PR's investigation (`DV_SLAM_LANDMARK_CAPTURE`) already merged in #190. The walker's same-colour soft penalty (W_SAME_COLOR_PENALTY = 1.5, from #187) stays — that's a real improvement that prefers cross-colour midpoints when both are available.

The real fix needs upstream work, not a planner-side patch. Documenting what we found in #189 for the next attempt.

Issue #177, opened 2026-05-01
slam: cone_graph_slam only maps ~7 cones from a stationary car at the start gate


raulmoranguerra:
## Symptom
With the car parked at the FS-DV start gate, \`cone_graph_slam\` reports \`map=7\` cones for steady-state \`step\` ticks (10, 20, ..., 90). The track has 168 cones and many are within visible range of a 360° / wide-FoV LiDAR mounted on the rollhoop. The Delaunay planner (and any planner that needs a forward corridor) starves on this cone count and returns empty paths, so the autonomy can't drive.

The deadlock: stationary car → SLAM only sees a handful of nearby cones → planner returns empty → controller fails safe to zero throttle → car stays stationary → no new cones come into FoV → loop.

## What we know
- Track loaded: \`test_submodule.csv\` with 168 cones, \`car_aligned: true\`
- LiDAR is configured in settings.json as a Hesai-ATX-class sensor (128 ch, ±60° H FoV, 0.1 m range noise) — should detect well past 20 m
- \`Cone_Detection\` upstream of cone_graph_slam IS receiving LiDAR data (no \`no_cones\` warnings from Plan_Path callbacks at 9.7 Hz)
- SLAM is in \`SLAM_RUNNING\` state with clean bias estimate
- The map size stays at 7 across many ticks while the car is stationary

## Hypotheses
1. **Cone_Detection's cluster-size or range filter is too tight** — cones at >10 m get rejected even when the LiDAR sees them. Worth inspecting \`pipeline/slam/slam/two_phase_cone_detection.py\` thresholds.
2. **cone_graph_slam's landmark-init confidence threshold** is too high — it sees the cones but rejects them as low-confidence. Inspect the new-landmark gating in \`cone_graph_slam_node.py\`.
3. **LiDAR FoV is narrower than configured** — Cone_Detection only processes 60° forward, missing side cones at the start gate.
4. **Persistent / multi-hit threshold** — cones are added to the map only after N consistent observations, and at the start gate we get N-1.

## How to investigate
1. From a Foxglove session at PIE start, check \`/Conos_raw\` (raw cone detections) vs \`/Conos\` (mapped landmarks) — the gap shows what cone_graph_slam is rejecting.
2. \`docker logs ifssim-dv_pipeline_stack-1 | grep cone_graph_slam | head -30\` after a fresh start — count \`obs=\` vs \`new=\` in step logs.
[...]

PR #195, opened 2026-05-02
chore(sim): correct LiDAR channel count to match Hesai ATX_S01 datasheet (128 → 116)

raulmoranguerra:
## Summary

The user shared the Hesai ATX_S01 user manual (Section 1.4: Specifications) and we verified the sim's Lidar1 settings against it:

| Parameter | Real ATX_S01 | sim (before) | sim (after) |
|---|---|---|---|
| Channels | 116 | 128 | **116** |
| H FoV | 120° (-60° to +60°) | 120° | 120° (unchanged) |
| V FoV | 18.3° (-12.4° to +5.9°) | 18.3° | 18.3° (unchanged) |
| Frame rate | 10 Hz | 10 Hz | 10 Hz (unchanged) |
| Points/sec | ~1.74 M @ 0.08° H | 200 k | 200 k (left as-is, see below) |
| Range precision | 3 cm 1σ | 3 cm | 3 cm (unchanged) |

The previous `_comment_lidar` in settings.json claimed 128 ch matched the real unit. The datasheet says otherwise. This PR corrects only the channel count and updates the comment to cite the datasheet.

## What about points/sec?

Sim runs at 11% of the real spec's points-per-second (200 k vs 1.74 M). Earlier history (in the original comment) shows raising it to 900 k previously CPU-bound the publish thread and dropped /lidar to ~5 Hz. The cone detector has been substantially revised since (#177 lowered cluster floors), so it likely handles sparser clusters now and the cap could be raised — but **changing one thing at a time**, the user explicitly wanted points/sec untouched until tested independently. Left at 200 k for this PR; comment tags it as worth revisiting.

## Test plan

- [x] JSON validates.
- [x] Restarted PIE + pipeline. Cone detection still produces healthy CONE_FILTER counts (11.4/scan accepted, no crashes).
- [x] Slight expected loss on long-range left cones (5.0 → 5.0... wait, actually 7.8 → 5.0); near-range and right side unchanged. No detection-pipeline errors.
- [ ] Drive a session and confirm no behavioural regression (the slight long-range loss is the expected datasheet-fidelity tradeoff).


raulmoranguerra:
Renaming branch — was 'chore', should be 'fix' (datasheet correction is a real bug fix). Reopening as fix/68-lidar-spec-from-datasheet, will also bump PointsPerSecond toward the datasheet's 1.74 M pts/s in the same PR.

Issue #199, opened 2026-05-02
perf: bridge UDP transport caps LiDAR pts/s ≪ datasheet's 1.74 M


raulmoranguerra:
## Symptom

When PointsPerSecond is set to the Hesai ATX_S01 datasheet value (1.74 M, 174 k pts/scan), the bridge enters a continuous reconnect loop:

```
[ifssim_bridge] WARN: LiDAR stream data incomplete
[ifssim_bridge] INFO: Reconnecting to IFSSIM (level reset?)...
[ifssim_bridge] INFO: LiDAR stream reconnected
[ifssim_bridge] WARN: LiDAR stream data incomplete
... (every ~10 ms)
```

The bridge does manage to deliver some points each cycle (we measured ~114 k of the 174 k actually arriving at Cone_Detection — a 35 % loss to fragment drops), but the reconnect churn is high.

## Why this matters

It's the bottleneck preventing the sim from running at the real Hesai's full pts/s rate. The Numba contiguous-array fix in #197 resolved Cone_Detection's CPU bottleneck — Cone_Detection itself can keep up with 1.74 M. The remaining limit is purely transport.

The undersampling at 200 k (11 % of real spec) means the sim's far-cone detection is artificially worse than the real car will see, which biases #189-style debugging toward solutions that match a sim quirk rather than real-car behaviour.

## What's happening

Each scan = 174 k pts × 12 bytes ≈ 2.1 MB. At 10 Hz that's ~21 MB/s sustained. Over UDP, with default datagram size of ~1500 bytes (Ethernet MTU), each scan fragments into ~1400+ datagrams. The bridge's reassembly window must catch every one of them; under load some fragments drop and the reassembly hits the "incomplete" path.

The bridge's UDP receiver code is in `ros2/src/ifssim_bridge/src/udp_receiver.cpp`. The FSDS plugin's UDP sender is in `Plugins/FSDSPlugin/Source/FSDSPlugin/Private/Networking/`.

## Candidate fixes (no investigation done yet)

1. **Bigger UDP socket buffers**. `SO_RCVBUF` / `SO_SNDBUF` defaults are typically 256 KB. Bumping to 16 MB on both sides would soak up larger bursts. `sysctl net.core.rmem_max` may need raising on Linux.
2. **TCP for the LiDAR stream**. Latency cost is low on localhost; reliability is the win. Other sensors stay on UDP.
[...]

PR #201, opened 2026-05-02
fix(bridge): TCP RCVBUF=16MB + xyz-only PointCloud2 + parallel ray traces (#199)

raulmoranguerra:
## Summary

Three fixes targeting the LiDAR throughput bottleneck identified in #199. Together they unlock 1 M pts/s (5× the previous 200 k baseline) with no disconnects, though the absolute publish rate at 1 M is bounded by TCP throughput on the test machine.

### 1. `SO_RCVBUF=16 MB` on bridge stream sockets

`IFSSIMRosWrapper::openStreamSocket` now sets a 16 MB receive buffer. Default (~256 KB) only held ~1/8 of a single 1 M-pt scan; under transient load the recv buffer would fill, the TCP window would close, the plugin's send-side flow control would stall, and the plugin's StreamLidar would tear down the connection ("LiDAR stream data incomplete" → reconnect loop).

### 2. xyz-only PointCloud2 with `memcpy` in `onLidarFrame`

The previous layout had a per-point FLOAT64 timestamp field added for fast_LIMO. fast_LIMO has been replaced by cone_graph_slam, which doesn't subscribe to `/lidar/Lidar1` at all — the only current consumer (Cone_Detection) reads xyz only. With a 3-field layout the whole copy collapses to a single `memcpy` of `total_points × 12` bytes.

### 3. `ParallelFor` over the horizontal trace loop in `FSDSLidarSensor::PerformScan`

`LineTraceSingleByChannel` is thread-safe on UWorld; each horizontal step runs on its own worker. `CurrentHorizontalAngle` was a per-iteration accumulator and is now derived from the h-index directly. Workers write to a pre-sized `SlotResults` array; a single-threaded post-pass packs hits into the output buffer.

## Verification at 1 M pts/s

| Metric | 200 k baseline | 1 M with all 3 fixes |
|---|---|---|
| pts/scan | 13 130 | 65 763 |
| clusters detected | 22 | 50 |
| accepted cones | 13.7 | 28.6 |
| by-side L / R / BO | 7.8 / 3.9 / 2 | 19.6 / 5.0 / 2 |
| `data incomplete` events | n/a | **0** |
| `/lidar/Lidar1` Hz | ~10 | ~3 |

Long-range left cones jumped from 7.8 → 19.6 per scan. Clean operation under autonomous driving (20 s, 50 m, 0 reconnects).

## Honest trade-off

The publish rate drops from ~10 Hz @ 200 k to ~3 Hz @ 1 M. Three independent bottlenecks were investigated:

1. **Bridge recv buffer** — fixed (RCVBUF=16 MB).
2. **Bridge publish thread** — fixed (xyz-only memcpy).
3. **UE5 ray-trace loop** — parallelised, but did not measurably improve the rate on this test machine.

The remaining limit is sustained TCP throughput (~3 MB/s observed on the LiDAR stream), and the 1.74 M datasheet rate still hits a plugin-side `SendAll` failure that needs separate investigation. Documented in `settings.json`.

Bridge-side fixes (1 + 2) are headroom-safe at any pts/s — no regression at 200 k. The ParallelFor (3) is worth keeping as future-proofing for higher-core-count machines and for when the transport bottleneck is addressed.

## Test plan

- [x] Container rebuild + UBT build succeed.
- [x] At 1 M: 0 reconnect events over 20 s autonomous drive.
- [x] CONE_FILTER counts confirm 2× more cones per scan.
- [x] At 200 k baseline: no regression.
- [ ] Real-car parity check on the eventual hardware.


raulmoranguerra:
Closing — superseded by later transport work on dev.

Issue #206, opened 2026-05-02
perf: tier-2/3 LiDAR scan-cost reductions (custom collision channel, async traces)


raulmoranguerra:
## Context

After the optimization pass in #72 (this branch), the plugin is at ~430 % CPU sustained on macOS Apple Silicon, with `FSDSLidarSensor::PerformScan` accounting for the bulk of it (174 k `LineTraceSingleByChannel` per second at the Hesai ATX_S01 datasheet rate).

The cheap fidelity-neutral wins are landed:
- UDP packet buffer hoisted out of the chunk loop (~2.5 k allocs/s saved).
- Per-channel direction vectors pre-computed once per scan (~348 k sin/cos calls/s → ~1.6 k).

What's left is structural and bigger.

## Tier 2 — custom LiDAR collision channel (most impact, low risk)

**Problem:** `LineTraceSingleByChannel(..., ECC_Visibility, ...)` traces against every Visibility-blocking actor in the level. UE5's broadphase walks the world spatial hash for each ray; even with `bTraceComplex=false` the per-ray overhead is substantial, and each call takes the Chaos physics scene read lock internally.

**Fix:**
- Define a custom `ECC_GameTraceChannel1` named `LiDAR` in `Config/DefaultEngine.ini`.
- Configure cone meshes, ground, and the car body to **block** the new channel; everything else **ignore** by default.
- Switch `PerformScan` to `LineTraceSingleByChannel(..., ECC_GameTraceChannel1, ...)`.

**Expected gain:** 30–50 % reduction in `PerformScan` CPU. The broadphase walks a much smaller spatial hash, and Chaos physics doesn't dispatch narrowphase against every irrelevant actor.

**Risk:** any actor not configured to block the LiDAR channel becomes invisible to it. Cones / ground / car must be audited; new map content needs the same configuration. Mitigation: a convention check at `BeginPlay` that warns when meshes labelled "Cone_*" don't have the LiDAR channel response set to `Block`.

## Tier 3 — async line traces (medium impact)

**Problem:** `LineTraceSingleByChannel` is synchronous — the calling thread waits for the trace result. With 4 worker threads in `ParallelFor` all hitting the Chaos read lock, traces serialise more than the worker count suggests.

**Fix:** use `World->AsyncLineTraceByChannel(EAsyncTraceType::Single, ...)` — queues into Chaos' async trace pipeline, results retrieved in a follow-up tick. Restructure `PerformScan` to be a two-phase issue/collect.

[...]

PR #214, opened 2026-05-02
feat(lidar): dedicated collision channel (tier 2 of #206)

raulmoranguerra:
First tier of [#206](https://github.com/isc-fs/IFSSIM/issues/206). Switches \`FSDSLidarSensor::PerformScan\` from tracing against \`ECC_Visibility\` to a dedicated \`ECC_GameTraceChannel1\` named \"Lidar\".

## Why

\`ECC_Visibility\` forces UE5's broadphase to walk every Visibility-blocking actor in the level per ray. With 174 k rays per scan that's the dominant cost in UE5's CPU profile. A dedicated channel that defaults to Ignore lets us narrow the candidate set to just the geometry the LiDAR cares about.

## How

Channel defined in \`Config/DefaultEngine.ini\`:
\`\`\`ini
[/Script/Engine.CollisionProfile]
+DefaultChannelResponses=(Channel=ECC_GameTraceChannel1,DefaultResponse=ECR_Ignore,bTraceType=True,bStaticObject=False,Name=\"Lidar\")
\`\`\`

Opt-in mechanisms (no per-asset editor work needed):
- **Cones** — \`FSDSConeSpawner\` sets \`ECR_Block\` on the new channel for every spawned cone (covers both the \`AStaticMeshActor\` path and the BP path).
- **Ground / landscape / static walls** — at \`LidarSensor::BeginPlay\` we walk the world once and promote anything that blocks \`ECC_WorldStatic\` to also block the new channel. Reliable proxy for \"things real LiDAR rays would hit\". Existing maps Just Work.

The car body is still excluded per-ray via \`TraceParams.AddIgnoredActor\` (unchanged).

## Honest performance result

Measured on macOS Apple Silicon, \`customMap\` (sparse geometry):

| Configuration | UE5 CPU |
|---|---|
| \`ECC_Visibility\` (baseline before this PR) | ~378 % |
| \`ECC_GameTraceChannel1\` (this PR) | ~370 % |
| Δ | **~8 %, within measurement noise** |

The savings are modest on this scene because the Visibility-blocker set was already small (~16 actors). On scenes with rich decoration, debris, or structures, the LiDAR channel filter would shrink the broadphase dramatically (Visibility candidate counts can run into the hundreds per ray on detailed maps). **The fix is structurally correct and its impact scales with scene complexity.**

The bigger near-term win on this sparse scene is **#206 tier 3 (async line traces)**, which addresses Chaos physics-scene lock contention rather than broadphase cardinality. Filing as a follow-up.

## Test plan

- [x] /lidar/Lidar1 publishes at 9.98 Hz, 119,445 pts/scan after the channel switch — no regression vs ECC_Visibility baseline.
- [x] World-walk log confirms 16 static-world actors auto-opted-in at BeginPlay.
- [x] Cones spawned via spawner are visible to the new channel (verified via probe count).
- [ ] On-track full-lap drive (waiting on PR #212 / issue #189 validation; merge order: #212 first, then this).


raulmoranguerra:
Reverting — channel switch broke cone visibility on the user's track. Investigation needed before retry: cones spawned before LidarSensor's BeginPlay world-walk, OR cone spawner's SetCollisionResponseToChannel call isn't reaching all cone primitives. Re-opening as a fresh branch once the opt-in timing is sorted.

PR #216, opened 2026-05-02
chore(settings): drop LiDAR rate 1.74 M → 1.0 M for sim CPU headroom

raulmoranguerra:
The datasheet rate (1.74 M pts/s) made the sim CPU-bound in PIE: Cone_Detection scaled linearly with point count and was hitting ~407 % CPU (≈4 cores) on its own, on top of UE5's PerformScan growing the same way. Combined Container + UE5 was running ~700 % CPU before the rest of the dev workflow even kicked in.

The real car runs the LiDAR on dedicated DSP hardware with no rendering / physics / SLAM-pipeline competition, so the datasheet rate is a non-issue there. In the sim every ray contends with the rest of the autonomy stack.

**1.0 M keeps the sensor saturating its 10 Hz publish budget cleanly** while freeing CPU for everything downstream. Wire rate lands around 650 k valid pts/s of returns (≈ 65 % hit ratio of attempted rays in an outdoor scene).

One-line config change — bumping back to 1.74 M is trivial when running benchmarks against real-car captures or validating sim/real parity.

## Test plan

- [x] /lidar/Lidar1 still publishes at ~10 Hz after Stop+Play (settings.json is re-read on PIE start).
- [ ] Cone_Detection CPU drops proportionally (~407 % → ~230 %).
- [ ] No autonomy regressions during a full lap.


raulmoranguerra:
Closing — `stat unit` showed the sim was GPU-bound (24 ms/frame), not LiDAR-bound. Lowering PointsPerSecond from 1.74 M to 1.0 M didn't move UE5 CPU at all. Real fix is the GPU side (#218). Keeping the LiDAR at the datasheet 1.74 M for sim/real parity.

PR #217, opened 2026-05-02
chore(lidar): wire MaxRange through settings.json (default 30 m for sim)

raulmoranguerra:
\`FFSDSSensorSettings\` already had a \`MaxRange\` field but it was never parsed from JSON or applied to \`LidarSensor\` — leaving the sensor stuck on its C++ default of 200 m regardless of what \`settings.json\` said.

## Wire-up

- \`FSDSSettings.cpp\`: parse \`\"MaxRange\"\` (meters, double) from the sensor JSON object.
- \`FSDSVehiclePawn.cpp\`: forward to \`LidarSensor->MaxRange\` (cm) with the standard m→cm × 100 conversion.
- \`settings.json\`: set \`MaxRange=30\` with a comment explaining the rationale.

## Why 30 m for the sim (not the datasheet 200 m)

The Hesai ATX_S01 spec is 200 m max range. For Formula Student cone detection nothing past ~30 m matters — the corridor width and the planner lookahead live well inside that. Shrinking the per-ray scan distance was the cheapest UE5 CPU optimisation we tried: Chaos walks the spatial hash for every actor along the ray's path, so a ~7× shorter ray cuts per-ray work proportionally in theory.

## Honest perf result

| | UE5 CPU | pts/scan | Hz |
|---|---|---|---|
| 200 m MaxRange (1.0 M PtsPerSec) | ~378 % | ~65 k | 10.00 |
| 30 m MaxRange (1.0 M PtsPerSec) | ~378 % | ~57 k | 10.00 |

Empirically the gain was modest — within measurement noise. The per-ray Chaos lock acquire + broadphase setup cost dominates over the spatial-hash traversal itself. Hit count dropped ~12 %, mostly from rays that previously found a distant ground hit.

Still worthwhile as config hygiene + small win + cleaner default for FS scenarios. Bumping back to 200 m is a one-line \`settings.json\` change.

## Test plan

- [x] /lidar/Lidar1 publishes at 10.00 Hz post-change.
- [x] pts/scan matches expected hit-rate for 30 m range.
- [x] No regression in cone visibility (cones are well inside 30 m).


raulmoranguerra:
Reopening with branch renamed to `fix/` per repo convention.

PR #218, opened 2026-05-02
chore(rendering): r.ScreenPercentage=25 to lift GPU bottleneck

raulmoranguerra:
\`stat unit\` showed the sim was GPU-bound, not CPU-bound:

\`\`\`
Frame: 24 ms  (~42 FPS, the bottleneck)
Game:   5 ms  (cheap — LiDAR ParallelFor lives in task graph workers, not the game thread)
RHI:  10 ms
GPU:  24 ms  (full frame budget burnt rendering the PIE viewport)
\`\`\`

The PIE viewport is purely a debug visual; physics, LiDAR ray traces, and the autonomy stack don't touch the rendered framebuffer. At 25 % render scale the GPU drops well under the frame-budget threshold so the editor stops dictating sim throughput.

Bump back to 100 in \`DefaultEngine.ini\` only when actually inspecting visuals.


raulmoranguerra:
Reopening with branch renamed to `fix/` per repo convention.

PR #264, opened 2026-05-04
fix(control): velocity setpoint + lookahead bugs on tight corners

raulmoranguerra:
## Summary

Three confirmed-bug fixes in the longitudinal/lateral controllers + a diagnostic tool, all motivated by the test_submodule first-hairpin runoff investigation.

After a deep audit, this PR was trimmed to *only* the changes with rock-solid evidence of fixing a real bug. Three previous commits were dropped: a `v_max` sledgehammer (was a wrong-hypothesis test), a cone-cull experiment that broke straight-line driving (immediately reverted in-PR), and an empty-chain-rejection fallback (real library bug but separate from controller fixes; reopen later if needed).

### 1. `fix(control): velocity setpoint now drops on tight corners`

Plumbs FaSTTUBe's analytical curvature through `Path → controller` so the longitudinal setpoint can actually see corners:
- `PathPoint` gains a `curvature` field; the adapter forwards FaSTTUBe's spline-derived κ verbatim.
- `path_planning._pose_stamped` smuggles κ via `pose.position.z` (flat-2D world; otherwise unused).
- `control_node._on_path` reads it back into `ReferenceTrajectory.curvature`.
- `_max_curvature_ahead` scans the full forward path (was windowed; missed corners on short paths).
- `a_lat_max` default 6.0 → 3.0. The old default was 0.6 g lateral — physically unrealistic for FS tires; v_corner = √(R · 6.0) was always above v_max for typical R, so the corner cap was non-binding.

Live verification: velocity overlay now shows `v_set = √(R · 3.0)` matches expected at every screenshot.

### 2. `fix(control): drop L_min refloor that defeated the hairpin radius cap`

Pure Pursuit's β·R radius cap (PR #261) was being silently overridden on hairpins tighter than `L_min/β`. With `L_min=1.5` and `β·R=1.4` on a 2 m hairpin, `Ld = max(min(velocity_law, β·R), L_min)` resolves to `L_min` — the cap is a no-op. Fix: drop the refloor, replace with a small absolute hard floor (0.5 m) so the chase target can't collapse onto the car.

`lookahead_min` default also lowered 1.5 → 1.0 to widen the band where the cap can bind.

### 3. `tools: offline replay tool for DV_PLANNER_CAPTURE JSONL dumps`

Adds `tools/replay_planner.py`. Plan_Path already supports per-tick scene capture via the `DV_PLANNER_CAPTURE` env var. This script replays a captured tick through FaSTTUBe and renders cones, pose, sorted chains, virtual cones, and centerline as a matplotlib PNG. Used during the audit to find the seed-starvation behavior in the library.

Runs inside the `dv_pipeline_stack` container where `fsd_path_planning` is installed:

```sh
docker compose exec dv_pipeline_stack \
    python3 /workspace/tools/replay_planner.py \
    /tmp/planner.jsonl --worst --out /tmp/replay.png
```

## Tests

- `pipeline/control/test/`: 7/7 pass (Pure Pursuit lookahead floor, radius cap, kappa helpers, edge cases).
- `pipeline/path_planning/test/`: 11/11 pass (planner adapter, cull, color routing, path arc cap, kappa pass-through).

## Test plan
- [ ] Drive test_submodule with the bind-mounted code; check velocity overlay confirms `v_set` drops smoothly into corners.
- [ ] Pair with PR #267 (orange-cone filter at SLAM input) for a clean cone map.


raulmoranguerra:
Velocity overlay from your last test_submodule hairpin run shows the v_set side of this PR is working correctly:

- `|κ| max ahead` = 0.503 /m → R ≈ 1.99 m
- `v_corner = sqrt(R · a_lat_max=3.0)` = 2.44 m/s
- `v setpoint` = 2.44 m/s ✓
- `v actual` (SLAM) tracking down to 2.27 m/s

Setpoint physics is healthy. **But the car still went wide** — root cause is on the Pure Pursuit side: the radius cap from #261 was being silently re-floored back to `lookahead_min`. With L_min=1.5 m and β·R=1.4 m, the cap was a no-op and Pure Pursuit chased a target *past* the apex.

Just pushed `3f9bb34`:
- Remove the `Ld = max(Ld, lookahead_min)` re-floor after the radius cap. New floor is an absolute 0.5 m so the chase target can't collapse onto the car.
- Lower `lookahead_min` default 1.5 → 1.0 (both the constructor and the ROS param in `control_node.py`).
- Update `test_lookahead_floors_at_lmin_on_extreme_curves` → `test_lookahead_stable_on_extreme_curves` and reword its docstring to match the new contract.

Tests still 7/7 green.

Re-test plan (same hairpin):
- `Plot!velocity` should show the same κ ramp / v_set drop as last run.
- 3D view: path tangent should track the corridor centerline through the apex without the chase target visibly leading past the inner cone line.

raulmoranguerra:
## Root cause found

I replayed `planner_capture_v5.jsonl` tick 530 (the last good plan before a previous runoff) through the live FaSTTUBe adapter using `tools/replay_planner.py`. Two side-by-side panels:

**Before** (`body_x > 0` cull active):
- 0 cones in front-left of the car
- Both side chains empty (`L=0 R=0`)
- Centerline is a 5 m stub anchored **30 m away from the car**

**After** (range-only cull):
- Both side chains populated (`L=11 R=11`)
- Centerline starts 3.3 m from the car and follows the corridor

The `body_x > 0` cull was the bug. It worked when the car sits near the corridor centerline, but at corner exits and anywhere SLAM hasn't yet incorporated upcoming cones, the cones the car needs to anchor the corridor topology are *behind* it. The cull dropped them and FaSTTUBe was left starved.

Across the previously-failing tick range 509–533, every single empty-path tick now returns a 24-point centerline. Adapter tests updated to pin the new contract (11/11 pass).

Pushed `f7a7c41`. Stack on this PR is now:
- `0afc49c` — curvature-adaptive lookahead + path arc cap (PR #261, already merged)
- `66bb44f` — diagnostic publishers (already merged)
- `85d448e` — analytical curvature plumbing + a_lat_max=3.0
- `3f9bb34` — drop L_min refloor on the radius cap
- `9e8cbbb` — v_max 3.0 → 2.0
- `3d185a3` — replay tool
- **`f7a7c41` — drop body_x>0 cone cull (this fix)**

Same retest as before (test_submodule first hairpin with overlay). If it makes it through, we can merge and look at backing v_max back up — speed was a coverup, not the actual problem.

Issue #269, opened 2026-05-04
SLAM diverges at second hairpin (DA wrong-match in cone-dense regions)


raulmoranguerra:
## Symptom

After PRs #264, #267, #268 land, the car successfully completes the first hairpin in test_submodule for the first time. Failure mode then shifts to the **second hairpin** (~tick 350 of a 364-tick capture, pose ≈ (20, 58, +173°)):

- yaw flips ~350° in a single tick (SLAM coordinate jump signature)
- pose-delta speed spikes to 11+ m/s (well above v_max=3.0)
- `||SLAM - GT||` spikes from a stable ~1 m for the whole run to 6.7 m at this point
- chains at the failing tick are populated with real cones (`L_real=8, R_real=8`) and the path looks geometrically correct

So this isn't a planner-side issue. SLAM lost localization mid-corner.

## Likely root cause

Even after the orange-observation filter (#267) and the planner colour-blind routing (#268), SLAM's **internal data association** still uses colour-gated matching at `pipeline/cone_slam/cone_slam/data_association.py` (`db.all_by_color(color)`). The upstream classifier mis-tags 30-50 % of cones, so DA gets fed wrong-colour candidates in cone-dense regions and Hungarian matches across mis-locked landmarks. iSAM2 snaps pose, cascade.

We attempted a fully colour-blind DA earlier and it caused 4.6 m drift in cluster regions (start gate, where orange and yellow cones sit within 2 m). So the fix needs to be smarter than dropping the gate.

## Proposed fix paths

**(a) Post-sort SLAM colour update.** Have the planner publish its `sorted_left`/`sorted_right` side-assignments back to SLAM, and use those as the colour-lock instead of the body_y heuristic. ~80-150 lines, all Python.

**(b) Tighter cross-colour DA gate.** Same-colour matches use full 2 m Euclidean gate; cross-colour matches use ~1 m.

**(c) Real LiDAR colour acquisition** (audit option a). Proper fix. Wire-format bump.

## Acceptance

Car completes ≥2 hairpins in test_submodule without `||SLAM-GT||` spikes above 2 m.

PR #279, opened 2026-05-04
feat(slam,path): planner-feedback colour update for SLAM landmarks

raulmoranguerra:
## Summary

The architectural fix from issue #269 option (b): SLAM stops trusting the unreliable `body_y` heuristic for landmark colour and instead consumes FaSTTUBe's geometric side assignments. The cross-colour DA gate (#272) and the pose-jump sanity check (#277) keep the residual flicker case recoverable while colours settle.

## Architecture

```
cone_slam --/Conos--> path_planning --[FaSTTUBe geometric sort]-->
  /path_planning/cones_left_ids   --> cone_slam: lm.color = BLUE
  /path_planning/cones_right_ids  --> cone_slam: lm.color = YELLOW
```

`/Conos` markers already carry the persistent landmark id as `marker.id`. `path_planning` now reads that id along with `(x, y, color)`. The FaSTTUBe adapter recovers per-side landmark IDs by matching `sorted_left` / `sorted_right` positions back to the input cone list (the library doesn't expose a direct index map; positions match within a 0.5 m gate, comfortably covering centroid noise).

## Why this should help where #277 fell short

The third-corner failure tail had the body_y classifier flipping a cone's colour mid-run as the car yawed through the corner. DA's cross-colour gate caught the flicker recovery, but with enough flicker, some bad matches still snuck through and snapped pose. FaSTTUBe's geometric sort doesn't depend on `body_y` at all — it classifies cones by which side of the corridor they're on, and that assignment is stable across yaw rotations. Once SLAM uses that as the colour source, the body_y mechanism that originally created bad matches is no longer in the loop.

## What changes

| Layer | Change |
|---|---|
| `core_types.Cone` | gains `id: int = -1` field |
| `path_planning._on_cones` | reads `marker.id` into `Cone.id` |
| `fasttube_adapter` | new helper `_match_positions_to_landmark_ids` recovers per-side ID arrays from `sorted_left/sorted_right` by nearest-neighbour position matching (0.5 m gate) |
| `PlanDebug` | exposes `left_landmark_ids`, `right_landmark_ids` |
| `Plan_Path` node | publishes `Int32MultiArray` on `/path_planning/cones_left_ids` and `/path_planning/cones_right_ids` every tick |
| `LandmarkDb.update_color(lid, color)` | new mutator; returns True iff landmark existed and colour changed |
| `cone_graph_slam_node` | subscribes to both ID topics; calls `update_color` with BLUE/YELLOW respectively |

Orange/big-orange landmarks are unaffected (they're already filtered at SLAM input per #267).

## Tests

- `test_landmark_color_update.py` (4 tests): basic update, unknown-id no-op, no-change return, persistence across `mark_observed`/`update_from_estimate`.
- `test_fasttube_adapter.test_plan_exposes_per_side_landmark_ids`: pins the per-side ID exposure on a clean straight track — both sides non-empty, every reported id is one of the input ids, sides are disjoint.

22/22 cone_slam + 12/12 path_planning + 7/7 control = **41/41 pass**.

## Test plan

- [x] `python3 -m pytest pipeline/{cone_slam,path_planning,control}/test/ -q`
- [ ] Live retest test_submodule. Watching for:
  - `/Conos` colours in Lichtblick look correct (no random orange/blue/yellow swaps as the car yaws)
  - `SLAM_OBS` log lines show fewer wrong-colour DA misses
  - Trajectory holds through the third corner (where #277's IMU-only recovery was masking the underlying classifier flicker)

## Closes

Implements option (b) for #269. The issue itself was already addressed by #272's cross-colour gate; this is the architectural fix that makes the gate unnecessary in steady state.


raulmoranguerra:
Closing — live retest showed the colour update creates obs/landmark colour disagreement that tightens the DA gate (cross-colour 1m vs same-colour 2m), causing earlier cascade than #277 alone. The architectural intuition was wrong because the body_y observation colour still flickers per scan even after landmark colour stabilizes. Will think about a different formulation.

PR #297, opened 2026-05-05
fix: move map→odom static from SLAM node to bridge — closes #295

### LINKED ISSUE #295: Lichtblick LiDAR cloud flickers off at session start (TF-tree gap)
## Symptom

On every session start (Play + Pipeline-Enable), the LiDAR cloud in Lichtblick briefly disappears in the world panel, then reappears together with the planner path. New observation, exposed once the LiDAR UDP wedge (#286) was fixed by host networking — until then, the cloud was usually invisible anyway.

## Root cause

`/tf_static` had two publishers:
1. `ifssim_bridge` publishing `base_link → fsds/{IMU,Lidar,GPS}` from container start.
2. `cone_graph_slam` publishing `map → odom` in its `__init__` — but only after Pipeline-Enable.

When (2) came online while a Lichtblick subscriber was already attached via foxglove WebSocket, the TRANSIENT_LOCAL handshake briefly recomputed the TF tree and the cloud lost its anchor for that window.

## Fix

Centralise all static TFs in the bridge. The `map → odom` identity moves from `cone_graph_slam_node.py` to `ifssim_ros_wrapper.cpp` next to the existing sensor identities, so `/tf_static` is fully populated at container start and stays bound for the container's lifetime. SLAM only publishes the dynamic `odom → base_link` afterwards.

Identity is still correct because we have no GPS-aligned global frame.

PR: #293.

raulmoranguerra:
Closes #295. Re-opens [#293](https://github.com/isc-fs/IFSSIM/pull/293), which auto-closed on the branch rename to follow `fix/<N>-...` naming.

## What

Move the `map → odom` identity static TF from `cone_graph_slam` into `ifssim_bridge`, alongside the existing `base_link → fsds/{IMU,Lidar,GPS}` identities. `cone_graph_slam` no longer owns any static TF — only the dynamic `odom → base_link`.

## Why

Symptom on `dev`: at every session start, the LiDAR cloud in Lichtblick briefly disappears, then reappears with the planner path overlaid. TF-tree gap caused by `/tf_static` having two publishers — the bridge (sensor TFs from container start) and the SLAM node (`map → odom` only after Pipeline-Enable). When the second publisher came online, Lichtblick re-handshook the new TRANSIENT_LOCAL backlog and the cloud lost its anchor for that window.

Was masked while the LiDAR UDP path was wedged (no cloud → no flicker visible). Now that #292 fixed the wedge, the flicker is exposed.

## Fix

Bridge owns all statics from container start. `/tf_static` is fully populated before the pipeline is even enabled. SLAM only publishes the dynamic `odom → base_link` and never touches `/tf_static`.

Identity is still correct — no GPS-aligned global frame; SLAM's `odom` is the `map` for downstream consumers.

## Verified live

After rebuild + restart with the pipeline disabled (SLAM not running):

```
$ ros2 topic echo --once --qos-durability transient_local /tf_static \
    | grep -E 'frame_id|child_frame_id'
    frame_id: base_link
  child_frame_id: fsds/IMU
    frame_id: base_link
  child_frame_id: fsds/Lidar
    frame_id: base_link
  child_frame_id: fsds/GPS
    frame_id: map
  child_frame_id: odom
```

Full chain present from t=0.

## Test plan

- [x] `pytest pipeline/cone_slam/test/` — 13 passed.
- [x] `/tf_static` shows full chain at bridge startup (above).
- [ ] Visual check in Lichtblick on next live session: cloud should stay anchored across Pipeline-Enable.


raulmoranguerra:
Closing per maintainer decision — using  as the Lichtblick display frame is a simpler workaround that sidesteps the flicker without code change. Bridge centralisation is still defensible architecture for future RViz / MCAP playback users; reopen if/when that surfaces. Branch left in place for reference.

Issue #301, opened 2026-05-05
SLAM cascade self-traps in FOV-limited corners → car off-track


raulmoranguerra:
## What goes wrong end-to-end on `test_submodule`

Live trace, 2026-05-05, `dev`-tip with all latest fixes (#290 z-anchor, #292 host networking, #298 startup steering ramp). SLAM tracks sub-meter (max 0.83 m, mean 0.59 m over 200 samples) for the first ~20 s of driving. Then the car goes off mid-lap. Sequence at the failure point:

```
travelled=72.5 m  v=3.88  steer=+0.634          # entering a right-hand bend
travelled=73.3 m  v=3.39  steer=+0.723

CONE_FILTER (avg/scan): pts=97k clusters=7.1 -> 5.1 fit
                        by-side: L=0.0 R=4.6 C=0.5 BO=0.0   ← only RIGHT-side cones visible

cone_graph_slam: skip cone factors: DA-failure spike (obs=14 new=11 assoc=3) — IMU-only update
cone_graph_slam: skip cone factors: DA-failure spike (obs=15 new=11 assoc=4)
cone_graph_slam: skip cone factors: DA-failure spike (obs=5  new=4  assoc=1)
cone_graph_slam: skip cone factors: DA-failure spike (obs=20 new=15 assoc=5)
cone_graph_slam: skip cone factors: DA-failure spike (obs=21 new=16 assoc=5)

travelled=87.6 m  v=2.77  steer=-0.798          # opposite-lock — controller chasing a wrong path
travelled=89.3 m  v=1.57  steer=-1.000          # full LEFT lock

cone_graph_slam: skip cone factors: DA-failure spike (obs=14 new=14 assoc=0)
cone_graph_slam: skip cone factors: DA-failure spike (obs=13 new=13 assoc=0)
... 15+ more "all-NEW" skips ...

travelled=96.2 m  v=3.57  steer=-0.076
travelled=109.2 m v=5.72  steer=-0.000          # 13 m of "travelled" in 0.5 s → pose dead-reckoning
```

## Failure mechanism

1. **Asymmetric cone visibility in mid-corner.** The car enters a section where one side of the cone corridor falls out of LiDAR FOV. Cone_Detection emits a scan with `by-side L=0 R=~5` (or vice-versa).
[...]

Issue #273, opened 2026-05-04
Cascade detector trips on legitimate transitions after #272 cross-colour DA


raulmoranguerra:
## Symptom

After merging PR #272 (cross-colour DA gate), the car completes both hairpins reliably but eventually loses SLAM localization on the third corner. Live test_submodule run got from "fails at first hairpin" to "completes both hairpins, fails at tick 410 on the third turn" — significant progress, new failure mode.

## Mechanism

The cascade detector in `pipeline/cone_slam/cone_slam/cone_graph_slam_node.py` rejects scans where the new-cone fraction looks suspicious:

```
WARN skip scan: DA-failure spike (obs=14 new=12 assoc=2) — pose-jump rejected
```

It was tuned for the colour-bucket DA, which produced near-100 % association rates on healthy track. With cross-colour DA, some legitimate transitions (e.g., a yaw rotation that flips colour tags on cones at the edge of perception) come back with mixed assoc/new counts that look cascade-shaped, but are actually just one-off noise. The detector skips those scans, SLAM's pose drifts during the skip (no constraints applied), and when the next valid scan arrives, the world has moved enough that *all* observations look new — at which point it really is a cascade, but a self-induced one.

Live trace from the run:
- step 380-390: `obs=13.8 (B=1.6 Y=12.2) assoc=12.6 (Y=11.0) new=1.2` — colour flipped, cross-colour DA recovered, healthy.
- step 400: `obs=13.8 (Y=13.8) assoc=9.2 new=1.8 skip=0.2` — first cascade-detector skip.
- step 401-410: more skips, drift accumulates during them.
- step 410+: `obs=13 new=13 assoc=0` — actual cascade after several skips.
- pose snaps from (10, 37) to (-7, 36) over 2 ticks → SLAM lost.

## Proposed fix

Loosen the cascade detector's threshold. Currently triggers at something like `new/obs > 50%`; should probably be `new/obs > 80%` or require N consecutive scans of high-new-ratio before rejecting.

Also worth investigating: skipped-scan handling. When a scan is skipped, SLAM should still update pose from IMU (which it does via preintegration) but ideally also run a "soft" update from the observed cones with looser constraints, rather than ignoring them entirely.

## Files

- `pipeline/cone_slam/cone_slam/cone_graph_slam_node.py` — cascade detector logic, ~30 lines around the `skip scan` warn.

## Acceptance

[...]

PR #303, opened 2026-05-05
fix(cone_slam): widen DA gate on cascade burst before falling back to skip — Lever 1 of #301

raulmoranguerra:
Implements Lever 1 of [#301](https://github.com/isc-fs/IFSSIM/issues/301). The full approach analysis is at [this comment](https://github.com/isc-fs/IFSSIM/issues/301#issuecomment-4381871923).

## What

When the cascade trigger fires (`obs ≥ 5`, `step > 30`, `>60 % NEW` at the default 1 m gate), the SLAM node now retries `associate()` at progressively wider Euclidean gates (3 m, then 5 m) before falling back to the IMU-only skip path.

If a retry drops the new-rate below 60 %, the wider-gate matches are accepted and cone factors commit normally — anchoring pose against existing landmarks instead of dead-reckoning.

## Why

`dev`-tip SLAM tracks **sub-meter (max 0.83 m, mean 0.59 m)** for the healthy section of `test_submodule`. The car still goes off mid-lap, every time. Live trace shows the failure mechanism:

1. Car enters a FOV-limited corner; Cone_Detection emits a one-sided scan (`by-side L=0 R=4.6`).
2. Predicted pose has ~1–2 m of cumulative IMU drift. The 1 m DA gate rejects the now-displaced observations as NEW.
3. Cascade detector fires → original recovery skips cone factors → IMU-only update.
4. Next scan: even more drift, more NEW. Cascade re-fires. Loop.
5. Plan_Path follows the diverging pose; controller commands oppose-lock; car off track.

The "skip cone factors" recovery was designed for #273 (joint-graph bias-feedback contamination), where committing the bad factors would corrupt B(k). On `dev` post-#290 (z-anchor), B(k) is still in the graph but the cone-factor → bias path is much weaker, and the cure (multi-scan IMU dead-reckoning) is now worse than the disease.

The wide-gate retry preserves the cone constraints and snaps pose back, which is what was needed all along.

## Safeguards

The original 1 m gate exists because cross-corridor mis-association at wider gates produced catastrophic divergence (#268, #272). To avoid re-introducing that:

1. **Single retry per cascade burst.** Once pose snaps back, the default 1 m gate resumes on the next scan. The Hungarian still enforces 1:1 matching across the wide gate.
2. **Widen-rate cap.** Track widen events; if ≥ 2 fire within 5 scans, escalate to lost-track instead of papering over. That signals the underlying cause (sustained drift, bias bug) isn't being addressed.
3. **Mahalanobis bypass on override.** The default sigma yields χ² ≈ 6 for a 1 m innovation, which would auto-reject every wide-gate candidate and defeat the retry. With a 5 m override, Mahalanobis is bypassed; Hungarian cost stays Euclidean.

## Test plan

- [x] `pytest pipeline/cone_slam/test/` — 15 passed (2 new tests on gate override behaviour).
- [x] Node imports cleanly with the new state.
- [ ] Live: complete a single lap of `test_submodule` (the bar that's never been cleared).
- [ ] If recurrent widen-rate-cap escalations to lost-track happen on the same corner, layer Lever 3 (asymmetric-scan landmark-creation gate) on top.

## Not in this PR

Levers 2 and 3 from #301 stay deferred. The bag-capture caveat (recorder suspected of contributing to crashes) means initial validation is live-only; if recording is workable, replay-harness regression follow-up via [#287](https://github.com/isc-fs/IFSSIM/pull/287).


raulmoranguerra:
Superseded by #fix/301-slam-cleanup-and-mahalanobis. The audit in [this comment](https://github.com/isc-fs/IFSSIM/issues/301#issuecomment-4383181387) traces what was actually pulling weight (the gate widen to 1.5 m) vs the layered-on band-aids (Lever 1 binary retry, smart-NEW phantom classifier, FAILED-widen cap, fall-through, sanity-check bypass) that introduced their own failure modes — the smart-NEW phantom feedback loop being the proximate cause of the post-second-hairpin runaway. Replacement PR replaces the lot with a single Mahalanobis adaptive gate. Branch left intact for diff reference.

Issue #298, opened 2026-05-05
Control saturates steering on first tick from planner's curved startup path


raulmoranguerra:
## Symptom

Every session start, control's first published `/control_command` has `steering=±1.0` (full lock). The car visibly yanks then self-corrects within ~1 second.

## Diagnostic (live, on dev)

Instrumented control's first 5 ticks. Tick #0:

```
pose=(0.022, 0.006, yaw=-0.0°)   # car at origin, facing forward
path_n=24
path[0]=(+0.19, -0.14)            # FIRST path point already 14 cm to the right
target_world=(+1.65, -0.52)       # 1 m lookahead is 52 cm right
target_body=(+1.63, -0.52)
Ld=1.02 m
kappa = 2 * 0.52 / (1.63² + 0.52²) = -0.36 → saturates pure pursuit
steering_norm=-1.000
cmd_steer=+1.000
```

Subsequent ticks (#1–#4) show the same pattern: planner's path[0] keeps being on the right by ~10–22 cm, lookahead 50 cm right, kappa saturates. The controller is doing exactly what pure pursuit must when the lookahead point is 18° off-axis at 1 m range.

## Root cause

The planner emits a **right-curving path from the very first frame**, before the car has moved. Likely FaSTTUBe's centerline computation from the first cone observations producing an asymmetric path (FOV-asymmetric cones, classification jitter, or virtual-cone insertion on a sparse side). Not a regression — this was masked by upstream issues (LiDAR wedge, SLAM crashes) before today.

## Proposed fix

Soft cap on `cmd.steering` magnitude during the first metre of travel. Linear ramp from ±0.3 at `travelled = 0` to ±1.0 at `travelled ≥ 1 m`. Doesn't hide bugs — the saturated value is a degenerate-geometry artifact, not a useful command — and the physical effect (car yanks at start) goes away.

Real fix would be on the planner side, but this is a 3-line band-aid and worth landing now.

Issue #286, opened 2026-05-04
Docker Desktop UDP proxy wedges port 51453 (LiDAR), survives container down/up


raulmoranguerra:
## Symptom

Mid-session, `/lidar/Lidar1` silently stops flowing into the pipeline. The bridge process inside the container looks healthy, foxglove keeps publishing, and `/imu`, `/motor_rpm`, `/testing_only/odom` all keep arriving. Only LiDAR drops.

On the host:

```
$ lsof -nP -iUDP | grep -E '51453|41452'
com.docke 30502 raulmoran 265u IPv4 ... UDP 127.0.0.1:41452
```

Port 51453 is missing from `lsof`, even though `docker compose ps` happily reports `127.0.0.1:51453->51453/udp` as a bound port.

## What does NOT recover it

Verified during 2026-05-04 session:

- `docker compose restart dv_pipeline_stack` — no rebind
- `docker compose up -d --force-recreate dv_pipeline_stack` — no rebind
- `docker compose down dv_pipeline_stack && docker compose up -d` — no rebind
- `tools/refresh-bridge.sh` (which already does the down/up sequence) — no rebind

After all of the above, the host still only shows port 41452. `com.docker.backend` has lost its grip on 51453 in a way that survives container teardown.

## What does recover it

Restarting Docker Desktop entirely. After Docker is back up, `tools/refresh-bridge.sh` rebinds both ports normally and LiDAR flows again on the next sim Stop/Play.

## When it triggers

Pattern observed across multiple sessions (cf. CLAUDE.md memory entry "Docker UDP proxy silently wedges"):

- High-bandwidth sustained UDP traffic on the port (LiDAR is ~10 Hz × ~80 KB).
- Repeated `compose restart` / `compose up --force-recreate` cycles in the same Docker Desktop lifetime.
- Coincides with the macOS networking stack going through a sleep/wake or VPN reconnect, but reproduces without those too.

## Workarounds (in order of preference)

1. **Restart Docker Desktop, then `tools/refresh-bridge.sh`.** Reliable; ~30 seconds of downtime.
2. **`LIDAR_TRANSPORT=tcp tools/refresh-bridge.sh`.** Bypasses the UDP proxy entirely. Costs throughput but avoids the wedge. Sim side has to be flipped to TCP too — there's a transport selector in the shipped build / DefaultGame.ini.
[...]

PR #285, opened 2026-05-04
fix(cone_slam): decouple IMU bias from cone factor graph (#282)

raulmoranguerra:
Closes #282. Re-opens [#283](https://github.com/isc-fs/IFSSIM/pull/283), which auto-closed when the branch was renamed.

## Summary

Refactors the cone-graph SLAM to a **position-only** factor graph (X(k) + L(id) only) plus an external `BiasEKF`. iSAM2 no longer holds V(k) or B(k), and cone bearing-range factors can no longer reach the bias state through joint optimisation.

## Why

In the previous architecture, a wrong cone factor that survived the DA gate satisfied its residual partly by adjusting B(k). That contaminated future IMU predictions, fed back into more wrong DA, and cascaded. We saw this end-to-end on `test_submodule`: a single mis-DA at ~tick 460 of an otherwise-healthy lap-long run snapped pose, the cascade detector skipped subsequent scans, IMU dead-reckoning accumulated drift, and the run was unrecoverable.

Five prior PR layers (cross-colour DA gate, cascade detector + IMU-only commit, post-iSAM2 pose-jump sanity check, body_y classifier removal, etc.) were band-aids on the same architectural feedback path. Decoupling bias estimation cuts that path entirely.

This is the AMZ / KIT / QUTMS pattern (FS-DV literature survey 2026-05-04): the IMU+RPM filter owns velocity and bias; SLAM consumes scan-to-scan pose deltas only.

## Changes

- **`bias_ekf.py` (new)** — 6-DoF random-walk bias filter. Seeded from the static-window calibration; covariance grows by RW noise on every IMU sample; corrected by a scalar Kalman update on accel-x from RPM-vs-IMU body-x velocity disagreement.
- **`factor_graph.py` (rewritten)** — drops `V(k)`, `B(k)`, `ImuFactor`, `BetweenFactorConstantBias`, `PriorFactorConstantBias`. Adds `stage_pose_delta(prev_pose, delta_pose, sigmas)` taking the IMU-derived `BetweenFactorPose3`. `commit()` now returns `gtsam.Pose3`; the node assembles a `ScanResult` separately.
- **`cone_graph_slam_node.py`** — wires `BiasEKF` through:
  - IMU callback: `predict_through(t)` advances bias covariance.
  - Cone callback: pulls latest bias into preintegrator, predicts a `NavState`, derives a body-frame pose delta, stages it as `BetweenFactorPose3`. After commit, runs an RPM-vs-IMU velocity comparison and feeds it back to the EKF.
[...]

PR #305, opened 2026-05-05
fix(cone_slam): cleanup band-aids + enable Mahalanobis DA — closes #301

### LINKED ISSUE #301: SLAM cascade self-traps in FOV-limited corners → car off-track

raulmoranguerra:
Closes #301. Replaces #303.

## Audit + decision

Full audit at [issue #301 comment](https://github.com/isc-fs/IFSSIM/issues/301#issuecomment-4383181387). Headline:

| Real contributors (kept on `dev`) | Removed band-aids |
| --- | --- |
| z-anchor (#290) | Cascade detector + IMU-only (#274) |
| Host networking (#292) | Pose-jump sanity check (#277) |
| Replay harness (#287) | Smart-NEW phantom classifier |
| Startup steering ramp (#298) | Lever-1/2/3a binary widen |
| | FAILED-widen cap |

The smart-NEW phantom classifier (PR #303) was the proximate cause of the post-second-hairpin runaway: 76 phantom landmarks created in 10 scans → pose teleported → controller chased fictitious velocity. Single-scan 5 m radius can't distinguish "obs is genuinely new (unmapped)" from "pose is wrong by >5 m". Removed.

## What's in this PR

**Removed:**
- `commit_with_pose_sanity_check` from `factor_graph.py` (the #277 sanity check).
- Cascade-trigger detection + IMU-only fallback from `_on_cones`.
- `pose_jump_max_pos_m` / `pose_jump_max_yaw_rad` param declarations.
- `test_pose_sanity_check.py` (covered the removed method).
- `skipped` per-scan diagnostic counter.

**Changed:**
- `DISTANCE_GATE_M` 1.0 → 1.5 m. Sub-meter SLAM on the post-#286 stack means the 1.0 m gate was rejecting legitimate matches as soon as corner drift hit the same magnitude as the gate radius. 1.5 m is still half the cross-corridor cone spacing.
- `_on_cones` now passes `pose_xy_yaw_cov` (extracted from iSAM2's 6×6 Pose3 marginal at indices [2, 3, 4] = (rz, tx, ty)) and `landmark_covariance_fn=self._graph.landmark_covariance` to `associate()`. The χ² gate adapts continuously to actual covariance — tight on healthy straights, wider during high-curvature manoeuvres where iSAM2's pose marginal grows.

The `POSE_COV_INFLATION` (16×) and `COV_FLOOR_VAR_M2` (0.49 m²) constants in `data_association.py` from the 2026-04-29 experiment remain. Those are the right knobs to tune live if the gate is too tight or too loose.

## Why this might work where 2026-04-29 Mahalanobis cascaded

Old stack: body_y classifier polluting matches, 1.0 m gate, LiDAR wedging, no z-anchor (so iSAM2 landmark marginals could go ill-conditioned via rank-1 deficiency on landmark.z).

Today's stack: classifier gone (#281), LiDAR reliable (#292), z-anchor in place (#290), 1.5 m gate. iSAM2 marginals are now usable as DA covariance estimates.

## Diff stats

```
4 files changed, 63 insertions(+), 338 deletions(-)
```

## Test plan

- [x] `pytest pipeline/cone_slam/test/` — 7 passed (2 new gate-override tests dropped with the levers; the rest are the default-gate DA contract).
- [x] Module imports cleanly.
- [ ] Live: complete a lap on `test_submodule`. The bar that's never been cleared.

## Out of scope

Control-side `a_lat_max=1.5` and `deadband=0.05` (from PR #303 commits) are independent improvements — separate small PR.


raulmoranguerra:
Closing — Mahalanobis adaptive gating didn't deliver. Live trace showed the pose-jump sanity check (necessary as last-resort guard against iSAM2 teleporting) firing every single scan, which is just IMU dead-reckoning by another name.

We've spent significant time iterating SLAM (PRs #285, #303, #305) without ever isolating whether SLAM is even the bottleneck. Next move: run a diagnostic experiment with control subscribing to /testing_only/odom directly. If the car laps with perfect pose, SLAM is the issue we need to solve. If it doesn't, we've been chasing the wrong layer.

Branch left in place for reference.

Issue #311, opened 2026-05-06
Pipeline freeze: state, structural ceiling, and forward paths for handoff


raulmoranguerra:
## Why this issue exists

The IFSSIM pipeline (cone_slam, control, path_planning, cone_detection) is being frozen and will be picked up by a new team in a separate repo, eventually loaded back into this monorepo as a submodule. This is the single document the new team should read first.

It captures (a) what works in the current pipeline, (b) what doesn't and why we couldn't fix it in this repo, (c) the structural ceiling we hit, (d) the diagnostic infrastructure left behind, and (e) three plausible forward paths.

The freeze does NOT include the runtime / bridge / sim plumbing — that stays under active development here. Only the four pipeline packages above are frozen.

---

### ✅ `cone_detection` — works
- Two-phase clustering (ground removal + EM cluster fit) on the Hesai-equivalent LiDAR scan.
- Sustains 14–22 cones/scan with healthy left/right split through both hairpins of `test_submodule`.
- Per-cone σ_xy is reported and propagated end-to-end.
- `bReturnPhysicalMaterial=false` in the UE5 LiDAR raycast means there's no real colour signal arriving — the body_y-sign+height classifier (PR #281) was rightly removed as cargo-cult heuristic. That's the upstream root of the SLAM problem below; see #255.

### ✅ `path_planning` — works
- FaSTTUBe drop-in (PR #244). Outputs a 24-point, ~12 m path at ~10 Hz with no `plan_empty` and no `tf_miss` over a full lap.
- Per-pose curvature published via `pose.position.z` side channel for the controller.
- CPU cost is high under macOS Docker emulation (#246 still open) but does not stall the pipeline.

### ✅ `control` — works (with smoothing fix in #308)
- Pure Pursuit + PI-velocity strategy stack, both behind ABCs. Swappable.
- Adaptive lookahead with β·R radius cap (#260 follow-up).
- Curvature-aware speed setpoint, kinematic-stop cap, big-orange stop-latch.
- Startup steering ramp (#298 / PR #300) absorbs the first-metre planner spike.
- **PR #308** adds throttle/regen/steering rate-limiting at the publish boundary. Rate-limit half of #151 lands here; first-order lag is left as optional.
- Open weakness: PI gains and lookahead constants are sim-tuned guesses, not modelled from real IFS-08 dynamics. Re-tune is owed.

### ❌ `cone_slam` — works on straights, fails on curves
[...]

Issue #306, opened 2026-05-05
GT-as-SLAM diagnostic: controller is a co-bottleneck, not just SLAM


raulmoranguerra:
## TL;DR

Today (2026-05-06) we ran a **GT-as-SLAM** diagnostic — replaced `cone_graph_slam` with a small relay (`pipeline/cone_slam/scripts/gt_pose_relay.py`) that publishes the sim's ground-truth pose under the same node-name + topic contract. The car received **perfect pose** through `/cone_slam/state` and `/tf`.

**Result:** the car completed a lap on `test_submodule`, but **hit cones, briefly went off track, then recovered**. With perfect pose. The controller / planner stack is not innocent — it's a real co-bottleneck, not a passive consumer that we can ignore while we fix SLAM.

This reframes the analysis in #301 (\"FOV-limited corners trip DA cascade\"). #301 is not wrong, but it's incomplete: there are **two** coupled failure modes, and we'd been treating it as one.

## Evidence

Diagnostic node: `pipeline/cone_slam/scripts/gt_pose_relay.py` (kept on disk — permanent triage tool, never deploy on real car).

To run: edit `docker/dv_pipeline_stack/pipeline_only.launch.py` to swap the `cone_graph_slam` `Node(...)` entry with `ExecuteProcess(cmd=['python3', '/dv_pipeline_stack_ws/src/cone_slam/scripts/gt_pose_relay.py'], output='screen')`, copy launch into the container, drive Mission Control as normal.

Key observation 1 — bridge does not fill the GT odom twist:

\`\`\`
ros2 topic echo --once /testing_only/odom
twist:
  twist:
    linear:  { x: 0.0, y: 0.0, z: 0.0 }
\`\`\`

The relay therefore finite-differences pose to derive body-frame velocity. Without that fix, Control reads `v=0`, floors throttle (\"v=0.00 -> thr=+0.600\") and saturates steering against a path it can't move along. (This bites any future consumer that relies on `/testing_only/odom` twist; worth its own clean-up.)

Key observation 2 — with the fix, with perfect pose, the controller logs during the lap:

\`\`\`
v=4.17 -> thr=+0.000 regen=+0.549 steer=+0.148
v=3.37 -> thr=+0.000 regen=+0.000 steer=-0.069
v=1.98 -> thr=+0.534 regen=+0.000 steer=-0.061
v=4.20 -> thr=+0.000 regen=+0.578 steer=+0.490
[...]

Issue #255, opened 2026-05-04
feat(lidar): per-point intensity in sim + intensity-based cone colour classifier


raulmoranguerra:
## Why
Today's cone colour classification path is geometric: `cone_slam/color_classifier.py` decides BLUE / YELLOW / ORANGE / BIG_ORANGE from the cone's body-frame y position (`±0.5 m` centre band → ORANGE). This breaks at tight corners where inner-side cones cross the centre band — they get permanently locked ORANGE in the persistent landmark map. Worked around in `fix/254` by mapping ORANGE→UNKNOWN so the FaSTTUBe planner ignores the bad colour signal, but that only covers up the symptom; we still don't have a real colour signal for downstream consumers (visualization, future cone-aware controllers, lap counting via big-orange).

A robust real-world solution is **LiDAR intensity-based classification**, used by AMZ Driverless, FaSTTUBe, GreenTeam Stuttgart, and Munich Motorsport (see published FS Driverless papers). Cones reflect at NIR (905 nm) very differently:

- FS blue: ~10–20% reflectance
- FS yellow: ~40–60%
- Orange: ~60–70%
- White reflective stripe (every cone): ~90%+

The reliable signal is the **bimodal intensity histogram per cluster** — the white-stripe peak is consistent across colours, and the body-only peak (lower mode) is the colour discriminator after range-normalisation.

## Real LiDAR (Hesai ATX-S01)

The ATX-S01 returns 8-bit per-point intensity (datasheet under `docs/ATX_S01_User_Manual_A07-en-250710.pdf`). The hardware path is ready; we'd just consume the intensity field.

## Sim plumbing

The current sim LiDAR (`#223` GPU path) captures `SCS_SceneDepth` only — depth, no intensity. The bridge currently emits `PointCloud2` with `xyz` only (deliberate per `feat/206` perf work). Three options to add intensity:

1. **Cheap (sim-only hack)**: post-decode lookup. `FSDSConeSpawner` already knows each cone's CSV-loaded colour. Build a `(world_xyz → cone_color)` index after spawn, and after the GPU decode pass, do a nearest-neighbor against it; map colour → fixed intensity LUT. Doesn't model range/angle falloff or the white stripe but is enough for testing the downstream classifier.
2. **Medium**: second SceneCapture pass with `SCS_BaseColor`. The decode shader samples both depth and base colour per ray; intensity = grayscale(base). Adds one render pass per scan.
[...]

Issue #173, opened 2026-05-01
Sim: integrate the FS-DV state machine as part of the car (mirror IFS08-DV-uDV micro-ROS firmware)


raulmoranguerra:
The FS-DV autonomous-system state machine (T 14.8.x: AS_Off / AS_Ready / AS_Driving / AS_Finished / AS_Emergency, with the ASB / SDC / R2D / RES handshakes) is implemented on the real IFS-08 as an MCU running micro-ROS. The firmware lives at [isc-fs/IFS08-DV-uDV (dev branch)](https://github.com/isc-fs/IFS08-DV-uDV/tree/dev).

In the sim, this is **part of the car**, not part of the autonomy stack. The sim has to model the MCU's behaviour faithfully so that:

- The autonomy stack interacts with the same state-machine surface it will see in production (via CAN-equivalent topics like \`/as_state\` / msg 0x502)
- Mission control (and any judge-style observer) sees the same timings / transitions / illumination signals the real car will exhibit
- A bug in the autonomy that violates an AS_state transition fails the same way in sim as it would on track

## Implementation options

**A. Reimplement in C++ inside the FSDSPlugin** — fastest, least faithful.
- New \`UFSDSStateMachine\` component on \`AFSDSVehiclePawn\` (or a new \`AFSDSDvECU\` actor).
- Subscribes to inputs from RPC + bridge: mission selection, RES Go signal, ASMS state, vehicle speed, EBS pressure (or sim equivalent), TS active.
- Implements transitions per Figure 15.
- Enforces T 14.8.4 (≥5 s AS_Ready before R2D) and T 14.8.5 (≥3 s AS_Driving stationary before motion).
- Publishes AS_state via a new RPC method (\`getAsState\`) and on a CAN-style UDP broadcaster topic so the bridge can republish as ROS \`/as_state\` (or msg 0x502).
- ASSI illumination (T 14.9): expose as a sim-side mesh / material parameter the pawn updates.
- Handles AS_Finished (mission complete + standstill ≥3 s) by reading from referee state.
- Handles AS_Emergency (SDC open at RES → distinct state with manual reset path).

Pros: a few hundred lines of C++, fully self-contained, no extra build dep.
Cons: drifts from the production firmware as the real-car repo evolves. Two implementations to maintain.

**B. Run the actual IFS08-DV-uDV micro-ROS firmware in a Docker sidecar** — most faithful.
[...]

PR #312, opened 2026-05-06
docs: integrated architecture for DV-pipeline-submodule handoff

raulmoranguerra:
## Summary

- New doc: \`docs/integrated_architecture.md\` — the target end-state architecture after the DV pipeline migrates to a separate team's repo and is attached here as a submodule.
- Splices the new team's autonomy-block diagram into IFSSIM's existing sim + bridge + mission-control-web + viz layers in a single Mermaid graph.
- Documents the integration contract (topic names, types, frames, quirks) and lists IFSSIM-side work owed for the migration.
- Adds a banner to the legacy \`autonomy_pipeline.md\` directing readers to the new doc so target-state and current-state aren't confused.

## Why now

The new pipeline team has produced their autonomy-block diagram. Their block stops at the FSDS topic boundary; ours owns everything below that boundary. The integration is the interesting part, and capturing it now (before code lands) makes the contract reviewable and avoids the typical \"discovered on integration day\" surprises.

## Concretely flagged for the migration

- Bridge topic prefixes: \`/imu\` → \`/fsds/imu\`, \`/gss\` → \`/fsds/gss\` (with type bump to \`TwistWithCovarianceStamped\`), \`/control_command\` → \`/fsds/control_command\`. Today's launch-time remaps go away.
- The \`/fsds/testing_only/odom\` empty-twist quirk that bit \`gt_pose_relay.py\` (PR #309) — same fix-or-workaround call lands on \`odometria_node\`.
- Mission Control web (FastAPI) transport options: \`rclpy\` action client to \`mission_control_node\` (recommended) vs full retire.
- Frame name: \`base_link\` → \`fsds/FSCar\` propagation through Lichtblick layouts and any frame-string assumptions.

## Test plan

- [ ] Mermaid renders on GitHub web UI.
- [ ] Cross-doc links resolve (\`autonomy_pipeline.md\` ↔ \`integrated_architecture.md\`, plus links to #311, #306, #173, #255, #309).
- [ ] Banner on legacy doc is the first thing a reader sees.

## Related

- #311 — pipeline freeze + handoff context (this doc is the architectural side; #311 is the inventory + structural-ceiling analysis).
- #173 — FS-DV state machine (subsumed by mission management on the new side).
- #309 / #308 — controller smoothing + GT relay PRs that just landed; both are referenced.


raulmoranguerra:
Updated the doc with the pipeline team's WhatsApp explanation. The corrections that materially change things:

1. **`sim_supervisor_node` simulates the IFS-08 micro (uDV)** — clarified explicitly. In sim it sources GO + sinks FINISHED; on the real car the physical micro plays that role.

2. **Control path corrected.** Previous revision drew `control_node → /fsds/control_command → bridge` directly. Actual flow is `control_node → mission_control_node → sim_supervisor_node → /fsds/control_command → bridge`, deliberately mirroring the real-car `DVPC → micro → CAN` chain so chain bugs surface in sim.

3. **Two-phase Action protocol** now documented as a first-class section. Phase 1 = startup (heartbeat + ready/failed + JIT warmup window for numba — explicit design fix for the FaSTTUBe-first-call-delay failure that bit our in-tree pipeline, our PR #300 workaround). Phase 2 = runtime (throttle / steering + emergency + finished).

4. **`odometria_node` removed** from the diagram and from the lifecycle list. The pipeline team confirmed it's being deleted. The question of where `/odom` comes from now is **Q1** in a new "Open questions" section because the answer materially affects whether the cone-only DA ceiling from #306 gets reintroduced.

5. **Mcbe transport simplified.** Only one viable shape now (rclpy Action client of `sim_supervisor_node`). The previous "Option A vs B" framing was wrong since mcbe targets the supervisor, not mission_control directly.

6. **"Open questions"** section added (Q1–Q6) so unresolved decisions are visible before integration kickoff: odometry source, gss type, DVPC-vs-supervisor in sim, frame name aliasing, twist fill, mcbe orchestration mechanism.

7. **Migration order** rewritten to start with Q1 + Q3 resolution and a runnable supervisor stub before the bridge rename PR triggers in-tree pipeline retirement.

Mermaid rebuilt with the corrected dataflow + bidirectional Action arrows for the two-phase protocol.

PR #318, opened 2026-05-08
fix(bridge): publish clean GT body-frame velocity on /testing_only/odom

raulmoranguerra:
Closes #315.

## Summary

\`/testing_only/odom\` shipped with default-zero \`twist\` because the odom publish block in the bridge filled only pose fields. Consumers expecting ground-truth velocity (\`gt_pose_relay.py\`, sim_supervisor's planned GT compare) had to finite-difference pose as a workaround. This PR wires a clean body-frame velocity field through the UDP \`SensorFrame\` and lets the bridge fill \`twist.linear\` directly.

## Changes (5 files, +46/-2)

**Plugin — wire format:**
- \`Plugins/.../FSDSUdpBroadcaster.h\` — added three floats at the END of \`FFSDSSensorFrame\` (\`GtVelBodyX/Y/Z\`, m/s, body frame). Existing field offsets unchanged.
- \`Plugins/.../FSDSUdpBroadcaster.cpp\` — populated in \`PackSensorFrame\` from the same source as pose: \`VehiclePawn->GetVelocity() * 0.01f\` (cm/s → m/s) rotated into body frame via the inverse of the pawn's quat, then UE5→ENU axis remap (negate Y) to match the existing GSS axis convention. No sensor in the loop, no noise.

**Bridge — wire format + publish:**
- \`ros2/.../udp_receiver.h\` — mirrored the three fields (\`gt_vel_body_x/y/z\`) at the same struct offset. Both \`#pragma pack(push, 1)\`; layouts must remain byte-identical.
- \`ros2/.../ifssim_ros_wrapper.cpp\` — odom publish block now sets \`twist.twist.linear.x/y/z\` from the new fields with a \`1e-9\` covariance diagonal (GT, but a finite floor so downstream consumers that weight by inverse variance don't divide by zero). The \`/gss\` block above is untouched — that topic keeps the noisy GSS sensor model.

**Docs:**
- \`docs/FUNCTIONALITIES.md\` — flipped the current-behaviour note from "asymmetric (clean pose / default-zero twist)" to the post-fix symmetric "clean GT pose AND clean GT twist." The integration contract in \`autonomy_pipeline.md\` already specified this target state and didn't need to change.

## Coordinated wire change

The \`SensorFrame\` struct grew by 12 bytes. The bridge's frame-size check (\`n == sizeof(frame) && magic == SENSOR_MAGIC\`) silently rejects packets that don't match, so:

[...]

PR #321, opened 2026-05-08
feat(lidar): per-point intensity following the Hesai ATX-S01 working principle

### LINKED ISSUE #255: feat(lidar): per-point intensity in sim + intensity-based cone colour classifier

raulmoranguerra:
Closes #255.

## Summary

The simulated LiDAR was XYZ-only on the wire. Real Hesai ATX-S01 returns 8-bit intensity per point following a well-defined physical model. This change implements the same model on the GPU LiDAR path so \`/lidar/Lidar1\` carries intensity that varies correctly with material reflectance, surface orientation, and range:

\`\`\`
intensity = ρ_905 × cos(θ_inc) × (R_ref / range)²
\`\`\`

## Implementation — three SceneCaptures share view geometry

| Capture | Source | Used for |
|---|---|---|
| Depth | \`SCS_SceneDepth\` (existing) | range \`r\` |
| BaseColor (new, #255) | \`SCS_BaseColor\` | ρ_905 placeholder via Rec.709 luminance |
| Normal (new, #255) | \`SCS_Normal\` | surface normal for \`cos(θ_inc)\` |

The decode compute shader (\`FSDSLidarDecode.usf\`) samples all three at the same texel as depth, computes intensity, and writes \`float4 (x, y, z, intensity)\` per ray. \`-1.0\` in w is the new invalid sentinel (distinct from a clean \`0.0\` intensity, which is a valid hit on a black surface or grazing-angle return).

The CPU fallback path computes the same formula using \`FHitResult::ImpactNormal\` and a placeholder ρ=0.5 (no per-material lookup on the CPU side).

## Wire-format coordinated change (per-point 12 → 16 bytes)

\`FFSDSLidarChunkHeader\` unchanged; only the per-point stride after the header changes. UDP chunk size constant 700 → 500 points/chunk so each 16-B/point datagram still fits under macOS's 9216 B cap (500×16 + 24 ≈ 8024 B).

**Plugin LiDAR senders patched on all three transports** — UDP, TCP, UDS — plus the \`getLidarDataBinary\` RPC reply, plus the \`OneScanMaxBytes\` SO_SNDBUF hint. **Bridge updates** in \`udp_receiver\`, \`tcp_client\`, \`ifssim_ros_wrapper\`. **PointCloud2 publish** now has 4 fields: \`x\`, \`y\`, \`z\`, \`intensity\` (FLOAT32 each); point_step 16.

## Coordinated rebuild required

Wire format grew 12 → 16 B/point so a bridge built from this PR running against an old plugin build silently rejects all LiDAR frames (frame-size check). Same discipline as #318: rebuild UE5 plugin + bridge together.

## Out of scope (follow-up)

- **Per-cone-material 905 nm reflectance tuning + white-stripe sub-material slot.** This PR uses \`luminance(BaseColor)\` as a placeholder for ρ_905. Tuning the actual Hesai datasheet reflectance values per cone-material (blue ≈ 0.15, yellow ≈ 0.50, orange ≈ 0.65, white-stripe ≈ 0.92) and adding a separate stripe material slot to get the bimodal intensity histogram pattern is a content task, deferred per discussion. The simulator-side principle is in this PR.
- **Hesai's auto-gain step** (low-priority; mostly matters at extreme ranges).

## Test plan

- [ ] \`tools/refresh-bridge.sh\` builds bridge cleanly. ✅ verified — container is healthy after the build.
- [ ] Rebuild UE5 plugin (Editor "Compile" / shipped build).
- [ ] Drive into a coloured cone; \`ros2 topic echo /lidar/Lidar1 --field intensity\` shows non-zero values that vary by surface and range.
- [ ] At 5 m and 15 m from the same cone, intensity ratio ≈ (15/5)² = 9× falloff (Lambert inverse-square).
- [ ] Surface normals matter: an off-axis hit on the cone produces a smaller intensity than a head-on hit at the same range.
- [ ] PointCloud2 in Foxglove/Lichtblick now exposes the \`intensity\` channel.

## Files changed

\`\`\`
Plugins/FSDSPlugin/Public/Sensors/FSDSLidarSensor.h          | +11
Plugins/FSDSPlugin/Private/Sensors/FSDSLidarSensor.cpp        | +153 / -··  (3 captures, shader bind, ConsumeReadbackResult, CPU fallback)
Plugins/FSDSPlugin/Private/Sensors/FSDSLidarDecodeShader.h    | +35  /  -·  (ColorTexture, NormalTexture, RotRows, RReferenceM)
Plugins/FSDSPlugin/Shaders/Private/FSDSLidarDecode.usf        | +77  / -··  (sample new textures, compute intensity)
Plugins/FSDSPlugin/Public/RPC/FSDSUdpBroadcaster.h            | +4
Plugins/FSDSPlugin/Private/RPC/FSDSUdpBroadcaster.cpp         | +15  / -··  (3 → 4 floats per point, 700 → 500 pts/chunk)
Plugins/FSDSPlugin/Private/RPC/FSDSRpcServer.cpp              | +22  / -··  (getLidarDataBinary, StreamLidar, StreamLidarUds)
ros2/src/ifssim_bridge/include/...                            |  -·       (no header changes — payload size is per-point param)
ros2/src/ifssim_bridge/src/udp_receiver.cpp                   | +22  / -··  (chunk math, assign size, 700 → 500)
ros2/src/ifssim_bridge/src/tcp_client.cpp                     | +2
ros2/src/ifssim_bridge/src/ifssim_ros_wrapper.cpp             | +50  / -··  (LiDAR receive, PointCloud2 publish 4 fields)
docs/FUNCTIONALITIES.md                                       | +20  / -··
docs/autonomy_pipeline.md                                     | +4   / -··
\`\`\`


raulmoranguerra:
## Soft-deprecation of TCP and UDS LiDAR transports

Pushed `11eb892` on this branch implementing the soft-deprecation we discussed:

- Bridge default `lidar_transport` flipped from "tcp" to "udp" (matches what `tools/refresh-bridge.sh` exports and what production runs).
- Bridge logs a deprecation WARN at startup whenever the resolved transport isn't "udp":
  > lidar_transport='tcp' is soft-deprecated (#321 follow-up). Production uses 'udp'. The 'tcp' path is kept for parity testing only and may be removed in a future release.
- `bridge.launch.py` env-var fallback flipped to "udp".
- `FUNCTIONALITIES.md` §6 reframed: section header now "UDP push, with TCP/UDS soft-deprecated"; transport table marks TCP/UDS with an asterisk and the WARN footnote.
- Verified: bridge restarts clean on the default UDP path with no deprecation WARN. No-op for production.

Tracking the **actual deletion** as #322 — purposely a separate PR so the wire-format-change review (this PR) and the cleanup-cost review don't conflate.

Issue #391, opened 2026-05-10
OdometryFilter: ZUPT (zero-velocity updates) + NHC (non-holonomic constraints) — bound vy + yaw drift during motion


raulmoranguerra:
**Two classic INS-aiding techniques to add to `OdometryFilter`** (`pipeline/sim_supervisor/sim_supervisor/odometry.py`). Both target drift modes that surfaced live during the #378 / #390 lap tests:

  • **Yaw drift +37° over 41 s of motion** (PR #378 lap analysis). Tracked indirectly as #386.
  • **vy accumulation to 3.22 m/s body-frame lateral** during autonomous driving (verified in #390 live test: `/odom.vy = 3.22` vs `GT.vy = -0.025`). The kinematic-bicycle `β_leak = 1e-3` is a weak form of NHC, but it's far too soft — accel_y from cornering integrates faster than the leak pulls back to zero.

## ZUPT — Zero-velocity updates

**Detection.** Vehicle is stationary when:
  • `|gyro|  < ω_threshold` (e.g. 0.02 rad/s) sustained for ≥ N samples
  • `|accel - g_body|  < a_threshold` (e.g. 0.05 m/s²) sustained for ≥ N samples
  • `|motor_rpm| < rpm_threshold` (e.g. 10 RPM) — drive shaft idle

**Update.** When detection fires, push two pseudo-measurements into the filter:
  1. `vx, vy = 0` (zero-velocity)
  2. Gradually pull `gyro_bias` toward the mean of recent gyro samples (since the car is truly stationary, any non-zero `gyro_z` is bias)

**Effect.** Every stop (pre-event, pit-in, between laps, controller-pause windows) hard-resets velocity to zero AND tightens the gyro bias estimate. Yaw drift gets bounded by the cumulative stop-time fraction. This is the canonical fix for the +0.9 °/s gyro bias residual #386 is tracking.

## NHC — Non-holonomic constraints

**Constraint.** For a ground vehicle in clean rolling, the body-frame lateral velocity is ≈ 0:
  • `vy_body ≈ 0` (no sideways slide)
  • `vz_body ≈ 0` (no flight)

These are *continuous* pseudo-measurements applied every IMU step, not just during stops.

**Update.** In a Kalman framework this would be `H = [0 1 0 ...]` (or whatever picks out vy) with variance σ²_NHC ≈ (0.05 m/s)². For the current complementary filter:

```python
# In push_imu, after the accel integration step:
[...]

PR #435, opened 2026-05-11
fix(control): state.speed = |vx| + matched Pure Pursuit lookahead retune (#424)

### LINKED ISSUE #424: Controller: bang-bang on tightest hairpin — tune Pure Pursuit lookahead + a_lat_max
## Observed
Post-#384 + #391 (ZUPT/NHC) live run on \`test_submodule.csv\`. The actuator chain held a clean closed loop for ~80 % of the lap (~187 m of clean cornering), then at the very tightest hairpin Pure Pursuit went bang-bang:

\`\`\`
travelled  v     steer    note
179.4m    2.99  +0.174   ← clean
180.5m    3.00  +0.098   ← clean
181.8m    2.99  +0.015   ← clean
183.2m    3.19  -0.134   ← clean transition
187.8m    2.97  -1.000   ← full lock — bang-bang re-emerges
190.4m    3.12  +0.966   ← snap opposite
193.8m    2.90  -1.000   ← snap back
197.7m    → off-line
\`\`\`

Path was stable through the entire window (\`path_n=24, path_len=11.6m\`); this is a controller-only failure, not a planner failure. Pre-#384 latency fix the bang-bang started at the second corner (~150 m); post-#384 it migrated to the very tightest one. That's consistent with the latency fix pushing the unstable-Ld point closer to the actual physical limit.

## Diagnosis
With \`lookahead_min=1.0\`, \`lookahead_k=0.5\`, at v=3 m/s the effective Pure Pursuit \`Ld = 2.5 m\`. \`test_submodule\`'s tightest corner has a local curvature radius < 2 m. When \`Ld\` exceeds the corner's natural lengthscale, the lookahead point jumps between adjacent path waypoints as the car traverses the corner — producing the ±1.0 steering snaps. \`a_lat_max=3.0\` (= ~0.3 g lateral) lets the longitudinal controller carry too much speed into the apex, compounding the lateral-loop instability.

### 1. Mechanical retuning (10 min, 90 % of the value)
- \`lookahead_min: 1.0 → 1.5\` — extends Ld at low speed so the controller sees further around the corner
- \`lookahead_k: 0.5 → 0.7\` — same for high speed
- \`a_lat_max: 3.0 → 2.0\` — drops corner-entry speed from v_corner = √(R · a_lat_max) so phase lag is less destabilising

These are \`declare_parameter\` calls in \`pipeline/control/control/control_node.py\`; override at launch time via \`ros2 param set\` or bake new defaults. Try one at a time, rerun \`test_submodule\`, log \`v_set\` + \`kappa_max\` + steering commands.

### 2. Curvature-aware lookahead (~50 lines)
Adapt \`Ld\` based on the local path curvature ahead of the car, not just velocity. Specifically: when \`max(|κ|)\` over the next 5 m of path exceeds a threshold, extend \`Ld\` linearly with κ. Counter-intuitively, a longer lookahead in tight corners produces SMOOTHER tracking — the controller targets a point further along the curve where the geometry is less locally extreme.

Implementation lives in \`pipeline/control/control/controllers/pure_pursuit.py\` — add a \`κ_max\` argument to \`compute()\`, extend the lookahead formula. Path's \`κ\` is already in the planner's output (per-pose, in \`pose.position.z\`).

### 3. Replace Pure Pursuit with LQR or model-predictive control
Architectural change. \`LateralController\` is already an ABC (\`pipeline/control/control/controllers/base.py\`); adding an LQR implementation is one new file + one branch in the factory in \`control_node._build_strategies\`. Heavier lift; only worth it if (1) and (2) don't get us a clean lap.

## Acceptance
- Clean lap on \`test_submodule.csv\` with no \`|steering| > 0.8\` in the last hairpin and no \`Δsteering > 1.0\` between adjacent 25 ms control ticks.
- Live run captured + analysed via the existing /control/v_set + /control/kappa_max diagnostic topics + the per-tick log line.

## Don't fix yet
- \`v_max=3.0\` is already very conservative. Don't lower further — the goal is to handle the corners we have, not slow the car down.
- Don't touch the planner. The diagnosis says path is stable; controller is the loop that's unstable.

## Related
- #384 (RuntimeControl actuator chain) — the latency fix that moved the failure point from corner 2 to corner 7
- #391 (ZUPT + NHC) — bounded the upstream pose noise; controller is now reacting to a clean path, so tuning will be reproducible
- #260 follow-up — \`v_set + kappa_max\` diagnostic topics already exposed for this analysis

raulmoranguerra:
Closes #424. **Two coordinated changes**; they have to ship together to keep the effective Pure Pursuit lookahead near its old "works on most corners" point.

## Root cause (from bag-replay analysis)

The supervisor's \`OdometryFilter\` integrates body-frame \`ay\` per IMU sample with only \`BETA_VY_LEAK=1e-3\` damping it. Centripetal \`ay = vx · ω\` is sustained during cornering — about 3 m/s² over a 5 s corner produces 15 m/s of integrated \`vy\` that the ~2.5 s leak can't bound. Today's bag (\`bags/lap_attempt_20260511_142317\`, 68 s autonomy window) shows:

- **Peak \`/odom.vy\` = 4.79 m/s**
- **Peak \`state.speed/|vx|\` = 2.59×** (\`vx=0.57\`, \`vy=-1.36\` → \`speed=1.48\`)

\`state.speed = √(vx²+vy²)\` folded that contamination into every consumer that read it:
- \`PIVelocity\` uses \`state.vx\` directly — unaffected.
- \`PurePursuit\` uses \`state.speed\` to size \`Ld = lookahead_min + lookahead_k · speed\`. Inflated speed → longer \`Ld\` → smoother corner tracking by accident. That accident is the **only** reason the prior runs reached travelled=187 m before bang-bang.

## What this PR does

| File | Change |
|---|---|
| \`control/state.py\` | \`speed = √(vx²+vy²) → abs(vx)\`. Controller stops being lied to. |
| \`control_node.py\` | \`lookahead_min: 1.0 → 1.2\`, \`lookahead_k: 0.5 → 0.6\`. Restores \`Ld=3.0 m\` at \`vx=3 m/s\` to match the pre-fix effective \`Ld\` (was 2.5 m post-fix-only). At \`vx=4 m/s\` we'd hit 3.6 m — the "cuts wide on corner 1" red line — but the longitudinal cap keeps us below that. |

## Trade-off

This does NOT fix the underlying \`/odom.vy\` drift. Any future consumer reading \`vy\` directly still inherits the contamination. The real fix is ZUPT+NHC (issue #391), done correctly this time. Tracked separately; this PR closes the controller-side blast radius now so a lap can complete.

## Verification

- [x] Container builds clean, both fixes present in installed source
- [x] Bag-replay analysis (\`bags/lap_attempt_20260511_142317\`): \`state.speed\` matches \`GT.vx\` within 0.05 m/s through the t=133–135 s corner window — controller now sees clean longitudinal velocity
- [ ] **Live trackdrive: expected to reach the previous farthest point (~167–187 m) without false PI-over-v_max trigger, with consistent throttle envelope through corners**
- [ ] If a regression: revert is two-line in \`state.py\` and four-digit edit in \`control_node.py\` — atomic, fast.

## Related

- Issue **#391** stays open — vy drift is upstream of this fix
- Issue **#424** closes via this PR
- PR **#433** (C++ filter, Phase 2) is orthogonal and still in flight


raulmoranguerra:
Closing — fix is technically correct (state.speed = |vx| is the honest reading; vy contamination peak dropped 4.79 → 2.41 m/s and inflation 2.59× → 1.39×) but is a **practical regression**: the post-fix bag (lap_attempt_20260511_144009_postfix) failed at corner 2, whereas the unfixed pre-fix baseline reaches the last hairpin at ~167-187 m.

Root cause: the pre-fix Pure Pursuit was *abusing* vy contamination as an accidental safety margin. Inflated state.speed → longer Ld in corners → Pure Pursuit survives tight turns. Removing the bug honestly, even with the matched lookahead retune in this PR, isn't enough to compensate. A constant lookahead bump has a narrow sweet-spot — 1.5/0.7 cuts wide on corner 1, 1.2/0.6 falls short on corner 2.

The real fix is curvature-adaptive lookahead (#424 option 2), not a constant lookahead bump. AND/OR fixing #391 properly so vy stays bounded at the source, which would then make this state.speed fix harmless.

Branch kept for resurrection: every line of code in here is still the right shape; it just can't ship alone.

Next session focus: fix the underlying /odom.vy drift (#391) via NHC done correctly (no slip-flag escape this time), then this PR can land cleanly because the controller won't have an accidental margin to lose.

Issue #315, opened 2026-05-08
bridge: route /testing_only/odom twist from clean kinematics, not from GSS


raulmoranguerra:
## Background

\`/fsds/testing_only/odom\` is documented (and used by consumers like \`gt_pose_relay.py\` and the planned \`sim_supervisor_node\` GT-compare) as a **ground-truth** odometry feed. Today it isn't, fully: pose is GT but twist is **left default-constructed (zero)**.

Verified in the source:

- \`ros2/src/ifssim_bridge/src/ifssim_ros_wrapper.cpp:832-851\` — the \`odom_pub_\` publish block fills \`pose\` from \`f.pos_x\` / \`f.pose_qx\` (clean GT — \`Frame.PosX\` / \`Frame.PoseOrient*\` come from \`VehiclePawn->GetActorLocation()\` and \`GetActorQuat()\` directly, no sensor in the loop). It does **not** fill \`twist\` at all — \`msg.twist\` is default-constructed and stays zero.
- \`Plugins/FSDSPlugin/Source/FSDSPlugin/Private/RPC/FSDSUdpBroadcaster.cpp:206-208\` — the broadcaster pulls \`Frame.GssVelX/Y/Z\` from \`VehiclePawn->GssSensor->GetOutput()\` (noisy — \`FSDSGssSensor.cpp:33-37\` adds \`VelocityNoiseStd\` Gaussian noise per axis), but those fields go only to \`/fsds/gss\` (lines 803-810 of the wrapper, inside the \`gss_pub_\` block). They are not connected to the odom topic.

Result: \`/testing_only/odom\` ships clean GT pose with zero twist. Anyone consuming the topic for ground-truth velocity comparison either gets misled (zero is "valid" if you believe it) or works around it locally (\`gt_pose_relay.py\` finite-differences pose for exactly this reason).

## Proposed change

Plugin side (\`FSDSUdpBroadcaster\`):

1. Add three clean body-frame velocity fields to \`FFSDSSensorFrame\` at the END of the struct (so existing offsets don't move):
   \`\`\`cpp
   // Ground-truth body-frame velocity (clean — no GSS noise) — #315
   float GtVelBodyX = 0.f, GtVelBodyY = 0.f, GtVelBodyZ = 0.f;  // m/s
   \`\`\`
2. In \`PackSensorFrame\`, populate them from the vehicle pawn directly (same source as pose):
   \`\`\`cpp
[...]

PR #443, opened 2026-05-11
fix(bridge): populate /testing_only/odom twist.angular.z with GT yaw rate

raulmoranguerra:
## Summary
- `GtVelBody*` (#315) closed the linear half of `/testing_only/odom`'s twist but left `twist.angular` at default zero, so consumers reading GT yaw rate saw a flat zero trace regardless of cornering. Confirmed empirically against `bags/lap_postveto_20260511_183638`: `n=9765 min=+0.0000 max=+0.0000 mean=+0.0000`.
- Add a companion `GtAngVelBodyZ` field on the wire (UE5 plugin → ROS bridge), sourced from the mesh's physics angular velocity rotated into body frame. Plugin-side applies the UE→ROS Z-axis pseudovector flip (same det(R)=-1 factor the bridge applies to `gyro_z` for `/imu`) so the bridge just copies into `twist.angular.z`.
- Both sender paths updated symmetrically: UDP (`FSDSUdpBroadcaster`) and TCP `streamSensors` (`FSDSRpcServer`).
- Refresh the stale comment in `gt_pose_relay.py` that still claimed the bridge ships `twist=0` — only `twist.angular` was zero before; now nothing is. The relay still finite-differences because its alignment is position-only, so the raw GT twist isn't directly reusable.

## Files changed
- **Wire format (both sides)**: `FSDSUdpBroadcaster.h` + `udp_receiver.h` — added a single `float` field at the end of the packed struct. UDP recv size-equality check stays satisfied; TCP `readExact` stays satisfied.
- **Plugin packing**: `FSDSUdpBroadcaster.cpp`, `FSDSRpcServer.cpp` — pull `GetPhysicsAngularVelocityInRadians()` off the skeletal mesh (same source the IMU sensor uses), rotate by the actor's inverse quat, write `-BodyAngVel.Z`.
- **Bridge publish**: `ifssim_ros_wrapper.cpp` — `twist.twist.angular.z = f.gt_ang_vel_body_z`, explicit `.x/.y = 0`, `twist.covariance[35] = 1e-9` (matches the GT linear floor).
- **Stale comment**: `pipeline/cone_slam/scripts/gt_pose_relay.py:84-89`.

## Rebuild status
- UE5 plugin: rebuilt locally via `Build.sh IFSSIMEditor Mac Development` — clean, 13 actions, `UnrealEditor-FSDSPlugin.dylib` linked.
- ROS bridge: `colcon build --packages-select ifssim_bridge` succeeds in the worktree's container (13.7s). The existing main `ifssim-dv_pipeline_stack-1` container still has the old build — run `tools/refresh-bridge.sh` after merge to pick up.
- Pre-existing nit: `tools/refresh-bridge.sh` ends by `docker compose cp`-ing `entrypoint.sh` into the new container, which doesn't preserve the exec bit (host file is `644`); the freshly-created container then can't start because `/entrypoint.sh` isn't executable. Hit it on the worktree-side stack — not in scope here. Existing main stack is unaffected because its entrypoint.sh was set executable at image-build time.

## Lichtblick
`lichtblick/clean_bag_replay.json` already has the GT yaw rate trace enabled on the yaw-rate plot at `#44ffaa` ([clean_bag_replay.json:183](lichtblick/clean_bag_replay.json#L183)) — once the new plugin + bridge run the trace will start showing real data, no JSON change needed.

## Test plan
- [ ] `tools/refresh-bridge.sh` from main after merge (and chmod the host entrypoint.sh if hit by the pre-existing nit above).
- [ ] Restart UE5 editor session, drive a cornering pass, record a fresh bag.
- [ ] Run the verification snippet from the bug report against the new bag — expect non-zero min/max on `/testing_only/odom.twist.twist.angular.z`.
- [ ] Cross-check against `/imu.angular_velocity.z` — same sign, same magnitude within bias-and-noise.
- [ ] Open the `clean_bag_replay` Lichtblick layout and confirm the GT yaw-rate trace tracks the IMU trace.


raulmoranguerra:
Superseded by #446. That PR has the same fix on a clean diff (6 files, 86 additions vs 48,830 here) — this branch was off a bad base and pulled in unrelated content/assets/configs. Closing in favour of the clean version.

PR #449, opened 2026-05-11
feat(filter): DV-PC EKF state estimator via robot_localization — Phase 1

raulmoranguerra:
**Draft** — opens Phase 1 of #447. Replaces the complementary \`OdometryFilter\` with a robot_localization-based EKF that lives on the DV-PC. The µDV is reframed as a sensor forwarder per the architectural decision in [#447 comment](https://github.com/isc-fs/IFSSIM/issues/447#issuecomment-4424150345).

## What this PR adds

| File | Role |
|---|---|
| \`docker/dv_pipeline_stack/Dockerfile\` | adds \`ros-humble-robot-localization\` |
| \`docker/dv_pipeline_stack/rpm_to_odom.py\` | adapter: \`/motor_rpm\` Float32 → \`/wheel_odom\` nav_msgs/Odometry, vx-only |
| \`docker/dv_pipeline_stack/ekf_dvpc.yaml\` | \`ekf_localization_node\` config: 2D mode, IMU predict, wheel_odom vx update, \`publish_tf=false\` for parallel-run |
| \`docker/dv_pipeline_stack/dvpc_ekf.launch.py\` | standalone launch, isolated from main pipeline |
| \`bags/compare_filters.py\` | replay harness — anchors GT, /odom, /odom_ekf to common start, prints residual envelope per filter + 5-s checkpoint table |

## What this PR does NOT do (yet)

- **Doesn't touch main pipeline.** The existing complementary \`OdometryFilter\` in \`sim_supervisor\` keeps owning \`/odom\` and the TF tree. The EKF runs alongside on \`/odom_ekf\`, observed-only.
- **Doesn't publish TF.** \`publish_tf: false\` in the YAML during validation. Flips to true in Phase 1.5 once we cut over.
- **Doesn't ship to µDV.** Phase 1 is DV-PC only. Whether \`libodometry_filter\` survives as a uDV-side library is a downstream decision.

## Verification path (merge gate)

1. Bring up the main pipeline as usual.
2. Bring up \`dvpc_ekf.launch.py\` alongside it.
3. Drive a lap, record a bag with \`ros2 bag record -s mcap -a -o lap_dvpc_ekf_<timestamp>\`.
4. Run \`bags/compare_filters.py <bag>\` — expect:
   - \`complementary\` residual: 15–35 m by t=80 s (as documented in #447).
   - \`EKF (robot_loc)\` residual: **sub-metre across the same window**.
[...]

Issue #447, opened 2026-05-11
OdometryFilter accumulates 15-35 m of position drift before SLAM cascade — root cause unknown


raulmoranguerra:
## TL;DR

`OdometryFilter` dead-reckoning is accumulating **15–35 m of position error before SLAM cascade onset** on a 60 s autonomy window. Visualised via a new Lichtblick gauge in `lichtblick/clean_bag_replay.json` that plots `||map → odom||` (the cumulative drift correction SLAM applies). On bag `lap_postveto_20260511_183638` this is the upstream timer that sets when the cascade detonates — it is *not* a SLAM-internal problem.

ZUPT + NHC (#391) were the canonical fix candidates but earlier experiments showed they made things **worse**, so we need to understand *why* the filter drifts before picking a remedy.

## Evidence

Plot of `||map → odom||` on `lap_postveto_20260511_183638` (post-veto bag, lap end at t≈83 s):

<img width="2168" height="1343" alt="Image" src="https://github.com/user-attachments/assets/9a5a9cb6-78f5-4cb4-871d-976fc93214cf" />

Key values pulled from the curve:

| t [s] | ‖map → odom‖ [m] | Phase |
|---|---|---|
| 0–22 | 0 | calibration / car stationary |
| 22 | ~3 (instant) | autonomy enabled — SLAM corrects 3 m on first scan |
| 22–45 | 3 → 7 | steady drift, ~0.17 m/s |
| 45–55 | 7 → 14 | first hairpin — drift jumps to ~0.7 m/s |
| 55–75 | ~10–15 (plateau) | straight section |
| 75–85 | climbs to 37 | second corner sequence |
| ≥ 85 | 67 → 95 (spike) | SLAM cascade detonates |

A healthy lap should keep this sub-metre for the whole 60+ s window. We are seeing **20-30× that** before SLAM even fails. SLAM was firefighting upstream drift the whole time; the proximity veto (#441) bought time but cannot fix the source.

## Why ZUPT + NHC isn't the answer (apparently)

Tried previously, made it worse. Worth unpacking *why* before retrying:

- **ZUPT** (zero-velocity update during standstill) is harmless when the car is genuinely still, but if the filter's "still" detector triggers during slow rolling, it injects a wrong constraint and can amplify bias-estimation error.
[...]

PR #451, opened 2026-05-11
feat(odometry_filter): 9-state EKF with bias states — Phase 2

raulmoranguerra:
**Draft** — Phase 2 of #447. Adds an Extended Kalman Filter sibling to the existing complementary \`OdometryFilter\`, with explicit bias states. Closes the bias-tracking gap that #447's investigation pinned as the root cause of 15-35 m of accumulated dead-reckoning drift per lap.

## State vector (9)

\`[x, y, yaw, vx, vy, omega, b_ax, b_ay, b_gz]\`

- **IMU** is the predict input — gyro_z drives omega, accel_x/y drive vx/vy, biases are subtracted and random-walked in Q.
- **/motor_rpm** is a vx measurement update on state[3].
- **NHC** (non-holonomic constraint) during clean-straight windows fires soft measurement updates of \`vy ≈ 0\` and \`omega ≈ 0\`, which flow through the state cross-covariance into \`b_ay\` and \`b_gz\`. This is the online bias-correction the complementary filter lacked.

All matrices are fixed-size Eigen — \`Matrix<double, 9, 9>\`, stack-allocated, no heap. The library stays MCU-portable even though the current architectural decision is uDV-as-sensor-forwarder.

## Bag-replay results vs the complementary filter (\`lap_postveto_20260511_183638\`)

| filter | mean | p95 | max |
|---|---|---|---|
| complementary (baseline) | 37.0 m | 76.0 | 76.1 |
| **EKF (Phase 2)** | **15.4 m** | **33.0** | **61.1** |

**2.4× reduction on the mean.** In the first 60 s of driving the EKF holds residuals to 2-17 m where the complementary climbs to 17-39 m — sometimes 10× better in early driving. Still not sub-metre — vy contamination during sustained cornering is the residual problem — but a clear directional win.

## What's in

| File | Role |
|---|---|
| \`include/odometry_filter/filter_interface.hpp\` | abstract \`IOdometryFilter\` (virtual push_imu/push_rpm/state/etc.) |
| \`include/odometry_filter/odometry_ekf.hpp\` | 9-state EKF class, Eigen fixed-size throughout |
| \`src/odometry_ekf.cpp\` | predict + scalar update math, calibration, brake-gating, kinematic-bicycle slip |
| \`include/odometry_filter/odometry_filter.hpp\` | \`OdometryFilter\` now derives from \`IOdometryFilter\` (interface only; math unchanged, all 17 existing gtests still pass) |
| \`test/test_odometry_ekf.cpp\` | 7 new gtests for the EKF |
| \`CMakeLists.txt\` | adds the new .cpp + test target |
| \`odometry_filter_node/src/odometry_filter_node.cpp\` | new \`filter_kind: complementary \| ekf\` parameter + \`ekf_enable_nhc\` for fine control |
| \`bags/compare_filters.py\` | anchor hardening — median over the first 0.5 s vs single-sample anchor |

## Tests

17 complementary + 7 EKF = **24 gtests, all passing.**

## Verification path

1. Bag-replay against \`lap_postveto_20260511_183638\` (this PR): EKF residual mean = 15.4 m, complementary = 37.0 m. ✓ directional win
2. Live-sim verification with \`filter_kind=ekf\` — pending, the merge gate.
3. Phase 2.5 — tune Q/R toward sub-metre, then flip the default \`filter_kind\` to \`ekf\`.

## Open issues

- EKF residual saturates at \`32.97 m\` from t≈95 s onward in the replay. Pattern matches a numerical instability in covariance late in the run. Worth investigating before merge.
- vy contamination during sustained cornering still leaks ~5-15 m of position drift over 60-90 s. The vy leak + NHC mitigate but don't eliminate. Either tighter \`R_nhc\` or a richer NHC trigger (e.g., averaged-velocity check) would tighten it further.
- Default remains \`filter_kind=complementary\` — flip in Phase 2.5 once we have sub-metre and a clean live-sim run.

Builds on #441 (proximity veto, merged), #433 (C++ filter wrapper, merged), #446 (bridge GT angular velocity, merged), supersedes the robot_localization experiment in #449 (kept as investigation tooling).

Closes (eventually) #447.


raulmoranguerra:
## Live verification finding: parallel-observer pattern destabilises live runs

Tested side-by-side today on identical track + driver:

| Run | Parallel EKF observer? | Autonomy length | Cascade at | Peak gt_err |
|---|---|---|---|---|
| \`lap_postveto_20260511_183638\` (baseline) | No | 63 s | t≈83 s | 1971 m |
| \`lap_ekf_rebuilt_20260511_232736\` | **Yes** | **34 s** | **t≈58 s** | cascade |
| \`lap_no_ekf_observer_20260511_234340\` | No | **64 s** | **t≈90 s** | 637 m |

Adding the parallel \`odometry_filter_node_ekf\` subscriber on /imu (333 Hz) + /motor_rpm + /steering_angle + /brake_pressure brought the cascade forward by ~25 seconds. With it removed, the system performs identically to baseline.

So **parallel-observed live testing isn't a valid verification path for Phase 2.** Two DDS subscribers on a 333 Hz topic in macOS Docker is enough to subtly starve the SLAM / control loop.

## Updated Phase 2 verification path

1. **Offline replay** is the primary validation gate — it's what \`bags/compare_filters.py\` already does, deterministic, no CPU contention. The 2.4× improvement we measured there stands.
2. **For live in-loop testing:** flip \`filter_kind: ekf\` on the main pipeline's \`odometry_filter_node\` instance directly. Only one filter publishing /odom + owning the TF tree. No parallel observer. Single-source A/B with one run per filter type.
3. The \`dvpc_ekf.launch.py\` pattern from #449 is fine for **offline-only** replays (no live pipeline pressure), but should not be brought up alongside a live lap again.

## What this means for the cascade root cause

The cascade in \`lap_no_ekf_observer\` (today's clean run) at t≈90 s has the same fingerprint as the cascade in \`lap_postveto\` at t≈83 s: gt_err sub-metre for 60+ seconds, then a sudden jump as pose drift past the DA gate spawns ghost cones. The proximity veto (#441) helps but doesn't eliminate this regime. The cascade timer is still set by **heading drift exceeding the veto threshold** — exactly what issue #447 documented and what the EKF's bias-state machinery is supposed to address. Need a fresh in-loop run with \`filter_kind=ekf\` (not parallel) to see whether the EKF's bias states bound heading well enough to delay or prevent the cascade.

This PR stays draft until that in-loop test happens.

raulmoranguerra:
## In-loop verification: three runs, definitive result

Same track, same driver, identical pipeline + cone_slam + planner + controller. Only filter swapped:

| Run | Filter (in-loop) | NHC | Autonomy length | Cascade onset | Pre-cascade gt_err |
|---|---|---|---|---|---|
| \`lap_no_ekf_observer_20260511_234340\` | complementary | n/a | **64 s** | t ≈ 90 s | < 1 m most of run |
| \`lap_ekf_inloop_20260511_235510\` | **EKF + NHC ON** | on | **64 s** | t ≈ 86–90 s | 1–6 m throughout |
| \`lap_ekf_inloop_nonhc_20260512_000116\` | **EKF, NHC OFF** | off | **32 s** | t ≈ 30 s of driving | 0.7 m clean → fast spike |

### Findings

1. **EKF + NHC ON ≈ complementary** in terms of when the cascade hits (~64 s both). The EKF's bias-state machinery doesn't measurably help in a single lap horizon — bias states don't have enough clean-straight time to converge.

2. **EKF without NHC cascades twice as fast** (~30 s vs ~64 s). The historical "NHC made things worse" pattern doesn't reproduce here — NHC is now load-bearing for in-loop stability, probably because the EKF's bias states absorb the side effects NHC would have created in the complementary filter.

3. **The cascade itself is gated by upstream drift exceeding the DA gate**, not by the filter quality. Different filters → same cascade timing. The filter isn't the binding constraint right now.

### What this means for #451

Phase 2 EKF as currently designed is a **correctness refactor** (proper Bayesian filter, explicit uncertainty, bias state vector), not an **autonomy win** in the 60-second-lap horizon. The 2.4× offline-replay improvement was real but doesn't transfer in-loop because:
- Offline replay processes deterministic, max-rate streams. In-loop has DDS jitter and callback contention.
- The lap is too short for bias states to converge from sustained clean straights.

### Recommendation

Keep this PR as draft / investigation. The autonomy fix lives elsewhere — a **count-based cascade detector in cone_graph_slam_node** that catches the 3+ ghosts-per-scan signature directly. Opening that as a separate PR now.

Until we have a long-lap evaluation (multi-lap trackdrive, real bias convergence window), this EKF doesn't ship as the default.

raulmoranguerra:
**Phase 2b just pushed** (commit `cbea5e6`) — drops the clean_straight gate, broadens both measurement updates.

### What's now in this branch

1. **NHC `vy ≈ 0`** fires every IMU tick post-calibration. R scales `(1 + |vx|/v_ref)` so the constraint self-loosens during cornering where real sideslip is expected.

2. **Kinematic bicycle `ω = (vx/L)·tan(δ)`** replaces the Phase 2a "NHC ω ≈ 0" gate. Implemented as a constraint-residual scalar update — `H = [0,0,0,-tan(δ)/L,0,+1,0,0,0]`, `z=0`. When δ=0 it reduces to the old ω≈0; when δ≠0 it predicts cornering yaw rate from steering + speed.

   Cross-covariance between `ω` and `b_gz` is what makes this load-bearing: the gyro-bias state now gets driving-time correction, the missing mechanism Phase 2a's calibration-only seeding couldn't provide.

   Gated on: `brake < kBrakeLockupThreshold` AND `|vx| ≥ 0.5 m/s`.

### Tests

3 new gtests, all pass:
- `NhcEnabledBoundsVyDuringCornering` — Phase 2a regression guard
- `KinematicBicyclePullsOmegaTowardSteeringPrediction` — primary contract
- `KinematicBicycleSkipsBelowMinVx` + `SkipsUnderBrake` — gates

Full suite: **11/11 EKF + 17/17 complementary filter tests pass**.

### Why the LWS sensor model (PR #470) matters

If we'd wired this against the pre-#470 perfect-echo `/steering_angle`, R → 0 and the EKF would over-trust δ during slip events. PR #470's quantized + nonlinearity-biased + hysteretic δ is exactly the right noise floor for `r_kb_omega = 0.05 rad/s`.

### Next step

In-loop verification — re-run the canonical lap with `filter_kind=ekf` and compare GT-integrated path-distance to cascade against the complementary baseline (~150 m on `track_20260404_013725.csv`, today's run). Expecting meaningfully further before cascade since Phase 2b adds the online gyro-bias correction that the in-loop tie at 64 s was diagnosing.

PR #453, opened 2026-05-11
fix(cone_slam): count-based cascade-spike detector

raulmoranguerra:
## Summary

Adds a second DA-failure-spike trigger inside \`cone_graph_slam_node._on_cones\`, AND'd with the existing post-discovery (\`step > 30\`) floor: **n_new ≥ 3 in a single mature-phase scan**. Catches the burst patterns that slipped through the legacy percentage gate (>60 %) and let cascades through in three separate runs today.

## Why

The historical 60 % gate missed the exact signature documented in #441's bag analysis:

> 14 observations, 6 flagged new (43 % — below the 60 % gate)
> 5 of those 6 had nearest-existing-landmark distance in [1.02, 1.20] m → cascade-defining ghosts, undetected

And the in-loop tests today (PR #451) showed the same pattern destabilising the autonomy regardless of whether the complementary filter or the new EKF was driving:

| Bag | Cascade onset | Filter |
|---|---|---|
| \`lap_postveto_20260511_142317\` | t ≈ 144 s | complementary (Python) |
| \`lap_no_ekf_observer_20260511_234340\` | t ≈ 90 s | complementary (C++) |
| \`lap_ekf_inloop_20260511_235510\` | t ≈ 86 s | EKF (C++) |

Cascade timing is set by drift exceeding the DA gate, not by filter quality. The filter is therefore not the binding constraint — a tighter cascade detector is.

FSD min cone spacing is 1.5 m (gates) / 3 m (track). At ~3 m/s a healthy scan adds at most ~1 new cone every 3-4 scans. Three or more new flags in a single mature-phase scan is anomalous regardless of total observation count.

## What this changes

| | Before | After |
|---|---|---|
| Module-level pure-function detector | inline in \`_on_cones\` | extracted as \`cascade_spike_triggered(n_new, total, step, ...)\` |
| Gate count | 1 (percentage) | 2 (percentage **AND** count) |
| Count-gate parameter | n/a | \`cascade_spike_new_count_threshold\` (default 3, 0 disables) |
| Trigger log | \`DA-failure spike\` | \`DA-failure spike [pct>60%, n_new≥3]\` — explicit which gate(s) fired |

## What this does NOT change

- Proximity veto (#441) — still active.
- Pose-jump sanity check — still active.
- Any filter (complementary or EKF) — untouched.
- Default count threshold = 3 is conservative; raise via parameter if a real lap pattern needs it.

## Tests

9 new pytest fixtures in \`test_cascade_spike_detector.py\` pin the gate behaviour at every interesting boundary:

- discovery-step floor (no trigger pre-step-30, fires at step 31)
- percentage gate fires at 62 % new, silent on small samples
- **count gate catches the 5-of-14 bag signature** (the bug #441 documented)
- count-threshold boundary (2 silent, 3 fires)
- count_threshold = 0 disables count gate cleanly
- both gates can co-fire and list both reasons
- healthy scan stays silent

All 9 new + 29 pre-existing cone_slam tests pass.

## Verification

- [x] Unit tests pass: \`pytest src/cone_slam/test/ -v\` — 38/38.
- [ ] **Bag-replay:** play \`lap_postveto_20260511_142317\` through the patched node, assert \`/Conos\` does not balloon at t ≈ 144 s (the cascade is caught and falls back to IMU-only update).
- [ ] **Live verification:** drive a fresh lap on TrainingMap with the patched detector, confirm the cascade is delayed past the historical t ≈ 60-90 s window. Will run as the next live test.

## Related

- Closes nothing yet — pending bag-replay + live verification.
- Built on top of #441 (proximity veto).
- Complementary to #451 (Phase 2 EKF, draft). The EKF is a correctness refactor; the count-gate is the autonomy fix.

The detector is the missing layer the cascade kept slipping through. Once landed, drift-driven ghosts get rejected before they can spawn new factors, regardless of whether the upstream filter is the complementary one or the EKF.


raulmoranguerra:
Live-test follow-up — assoc==0 narrowing pushed (88bf73c).

The first live verification went off on the first turn. slam_node logs showed the count gate firing on `obs=12 new=7 assoc=5` — legitimate cone discovery during cornering, not a cascade. Refusing those cone factors starved SLAM into IMU-only dead-reckoning.

**Fix**: count gate now requires `n_new ≥ threshold` **AND** `n_associated == 0`. The zero-assoc requirement discriminates a true cascade (pose off the map → nothing matches → all-new) from cornering discovery (some prior cones still associate). Percentage gate unchanged.

Replay of the bag signatures against the new logic:

| obs | new | assoc | label                | result      |
|----:|----:|------:|----------------------|-------------|
| 12  |  7  |  5    | legit turn discovery | silent ✓    |
| 12  |  9  |  3    | borderline           | pct gate ✓  |
| 11  | 11  |  0    | full cascade         | both ✓      |
|  9  |  9  |  0    | full cascade         | both ✓      |
|  4  |  4  |  0    | cascade tail         | count ✓     |

11/11 unit tests pass. Next: re-run the live verification lap.

PR #481, opened 2026-05-14
fix(docker): cross-platform compose — bridged net + explicit port forwards

raulmoranguerra:
## Summary

Switch `dv_pipeline_stack` from `network_mode: host` to **bridged networking with explicit port forwards**. One committed compose file works on Linux, Mac, and Windows with **no per-developer setup** (no override, no Docker Desktop toggle, no overlay).

Forwards added:
- `127.0.0.1:8765:8765/tcp` — foxglove WS (Lichtblick)
- `127.0.0.1:41452:41452/udp` — sensor UDP fanout
- `127.0.0.1:41500:41500/udp` — LiDAR chunked-UDP stream

Plus two stale `51453` → `41500` cleanup comments left over from #477, and a friendlier `.gitignore` comment for the now-rarely-needed `docker-compose.override.yml`.

## Why

Host networking has divergent behaviour across the three Docker Desktop backends:

| Path | Status |
|------|--------|
| Linux native Docker | works |
| Mac Docker Desktop ≥4.34 + "Use host networking" toggle ON | works (#286 verified) |
| Mac Docker Desktop with toggle OFF (fresh-install default) | broken — silent chunked-UDP drop |
| Windows Docker Desktop (WSL2 backend) | broken — chunked-UDP doesn't make it to Windows loopback regardless of toggle |

Bridged + explicit forwards sidesteps all three host-networking proxy paths. The original #286 userspace-UDP-proxy wedge that motivated host networking has been fixed upstream over the 6+ Docker Desktop minor releases since (move from vpnkit to gvisor in late 2025).

### Windows + Docker Desktop (WSL2 backend) — verified
- All four containers come up healthy with no override.
- `/imu` 443 Hz, `/gss` 110 Hz, `/gps` 11 Hz, `/testing_only/odom` 110 Hz ✓
- Lichtblick (http://localhost:8080) → `ws://localhost:8765` ✓
- Mission Control (http://localhost:3000 + :8000) ✓

### Mac — **pending verification** post Docker Desktop update
Expected to work; if #286's userspace-proxy wedge has been fixed upstream (likely with the gvisor migration), this PR is a complete cross-platform fix. Reviewer running this branch on Mac please drop a comment with `/lidar/Lidar1` rate from a sustained 30 s Play session.

### Linux — pending verification
Should be transparent (bridged is the default native Docker behaviour). Reviewer with a Linux Docker setup please verify.

## Known caveat — Windows LiDAR rcvbuf

`/lidar/Lidar1` on Windows occasionally goes silent under sustained chunked-LiDAR load. Diagnosed: WSL2 kernel UDP recv buffer fills faster than the bridge drains, `/proc/net/snmp` `UDP RcvbufErrors` counter climbs continuously. The bridge requests 16 MB via `setsockopt` but Linux honours up to `net.core.rmem_max` which inside the Docker Desktop VM is locked small.

This is **a pre-existing, separate issue independent of this PR's network_mode change**. It would happen under host networking too — Mac just has more headroom in its host kernel. Tracked as follow-up: either raise the VM's `rmem_max` via a Docker Desktop sysctl override, or move the LiDAR chunk pacer to drain-driven instead of fixed-interval. Not blocking this PR — non-LiDAR Windows development continues to work end-to-end.

## Test plan

- [x] Windows: stack up + sim launch + sensors at nominal rates + Lichtblick connects (above)
- [ ] Mac (post Docker Desktop update): /lidar/Lidar1 holds 10 Hz over 30 s play
- [ ] Linux: stack up + sensors at nominal rates
- [ ] Smoke that `tools/replay.sh` still works against `clean_drive_75s` (bag replay path is independent of network_mode but worth confirming)


raulmoranguerra:
Closing — verified live that this PR reintroduces #286. Synthetic UDP smoke test shows port 41500's userspace proxy binding silently dies under sustained chunked-LiDAR load (Docker Desktop's proxy hasn't been fixed upstream the way I assumed — the original #286 wedge mode is still very much alive). `network_mode: host` is doing real load-bearing work and shouldn't be removed without a transport-level solution.

See [discussion in chat] for the diagnostic trail. The real options forward remain (a) move LiDAR off chunked UDP onto the TCP RPC channel, or (b) ship Windows as bag-replay-only and keep `network_mode: host` as the committed default. Neither was done here.

Issue #322, opened 2026-05-08
Delete soft-deprecated TCP and UDS LiDAR transports


raulmoranguerra:
Follow-up to #321 — the simulator-side LiDAR pipeline runs on UDP push in production. The TCP and UDS LiDAR senders are kept as parity-test fallbacks for now, but every wire-format change has to touch all three sender sites in lock-step (#321 was the most recent example). The cost of maintaining that is no longer worth it.

PR #321's `11eb892` commit took the soft-deprecation step:

- Bridge default \`lidar_transport\` flipped from "tcp" to "udp".
- Bridge now logs a deprecation WARN whenever the resolved transport isn't "udp".
- \`bridge.launch.py\` env-var fallback flipped from "tcp" to "udp".
- \`FUNCTIONALITIES.md\` §6 reframed accordingly.

This issue tracks the **deletion** step — proceed once we've confirmed nothing off-Mac depends on TCP or UDS.

## What to delete

**Plugin side:**
- \`FFSDSRpcServer::StreamLidar\` (TCP) — the whole method body in \`FSDSRpcServer.cpp:1659+\`.
- \`FFSDSRpcServer::StreamLidarUds\` (UDS LiDAR) — \`FSDSRpcServer.cpp:1939+\` (the sensor UDS stub at \`StreamSensorsUds\` is separate, leave that alone).
- The \`streamLidar\` command branch in the RPC server's command dispatcher (\`FSDSRpcServer.cpp:309\` and the UDS dispatcher).
- The \`SetSendBufferSize\` block sized for one-scan TCP-LiDAR throughput (\`FSDSRpcServer.cpp:240-253\`) — was only there to soak up 3 cm range jitter on TCP.
- \`getLidarDataBinary\` RPC reply (\`FSDSRpcServer.cpp:1509+\`) — currently unused by the bridge or any documented client.

**Bridge side:**
- The TCP-LiDAR path in \`IFSSIMRosWrapper::lidarStreamThread\` (\`ifssim_ros_wrapper.cpp:560-600\` block) and any related connect/reconnect logic.
- The UDS branch in \`openStreamSocket\` (\`ifssim_ros_wrapper.cpp\` lines around the \`AF_UNIX\` connect attempt) — only active when \`command == "streamLidar"\` and \`lidar_uds_path_\` is set.
- The \`lidar_transport\` parameter in the bridge node — and the validation block at the top of the constructor.
- \`lidar_uds_path\` parameter in the bridge node.
[...]

PR #482, opened 2026-05-14
feat(lidar): TCP transport — closes the Docker Desktop UDP wedge

raulmoranguerra:
## Summary

Moves the LiDAR wire path from chunked UDP (introduced in #322) to a **streaming TCP push** on the existing RPC port (41451), mirroring the sensor stream's pattern. Solves two cross-platform wedge modes that the UDP path could not avoid:

| Wedge | Affects | What this PR does |
|-------|---------|-------------------|
| Docker Desktop userspace UDP proxy silently loses port bindings under sustained fragmented-UDP load (#286) | Mac (without host-networking toggle) AND Windows | Bypasses the UDP proxy — TCP uses a separate proxy path that's reliable |
| WSL2 kernel UDP `rcvbuf` overflow drops ~28 % of chunks | Windows | TCP is a byte stream — no recv buffer pressure from fragment bursts |

The #322 motivator (~7 MB/s macOS Docker Desktop TCP loopback cap) was fixed upstream over the 6+ Docker Desktop minor releases since then (gvisor proxy shipped late 2025).

## Wire format

```
[FFSDSLidarStreamHeader  24 B]   magic / frame_id / channels / total_points / lag_ns
[TotalPoints × 4 floats × 4 B]   (x, y, z, intensity)
```

One frame per scan. No chunking. `static_assert` on the 24-byte header size as ABI safety on both sides.

## What changes

**Sim plugin**
- `FFSDSRpcServer::StreamLidar(FSocket*)` — new streaming method paralleling `StreamSensors`. Polls `LidarSensor->GetTimestamp()`, emits a wire frame per scan until the bridge disconnects.
- Dispatch wired in `HandleClient` (command name: `streamLidar`).
- New on-wire struct `FFSDSLidarStreamHeader` in `FSDSUdpBroadcaster.h`.

**Bridge**
- `IFSSIMRosWrapper::lidarStreamThread()` — TCP recv loop. Reads stream header, validates magic + bounds, reads payload via `readExact`, hands off to the existing `lidar_pending_` single-slot buffer. Reuses a thread-local point buffer.
- `triggerReconnect()` now reopens both sensor + LiDAR TCP fds.
- Env-var escape hatch: `IFSSIM_LIDAR_TRANSPORT=udp` preserves the pre-PR chunked-UDP path. Default is TCP.
- UDP receiver + chunked-reassembly callback stay in the tree, used only on the `=udp` branch.

**Docker compose**
- No changes. TCP LiDAR rides the existing port-41451 forward. `network_mode: host` stays committed — orthogonal to this PR.

### Windows + Docker Desktop (WSL2 backend) — verified ✓
```
/imu                      406 Hz
/lidar/Lidar1             10.1–10.4 Hz over sustained 30 s
                          ~95k points/scan, std dev 40 ms
                          zero bridge warnings/errors
/gps                      10.8 Hz
/gss                      115 Hz
/testing_only/odom        108 Hz
/motor_rpm                106 Hz
/steering_angle           99.5 Hz
```

### Mac (pending)
Expected to work without the host-networking toggle. Reviewer with Docker Desktop ≥4.34, please:
1. `git checkout feat/lidar-tcp-transport`
2. `docker compose down && docker compose up -d --force-recreate`
3. Launch the sim, load a track
4. `docker exec ifssim-dv_pipeline_stack-1 bash -lc 'source /opt/ros/humble/setup.bash && source /dv_pipeline_stack_ws/install/setup.bash && ros2 topic hz /lidar/Lidar1'`
5. Expected: `~10 Hz` sustained, bridge log shows `LiDAR transport: TCP (streamLidar, PR-#482)`

### Linux (pending)
Expected to work transparently. Same test recipe.

## Known follow-ups (not blocking)

- Plugin-side `FSDSUdpBroadcaster::BroadcastLidarFrame` still sends UDP LiDAR unconditionally — those datagrams now hit a closed container port. Cosmetic; ~1.5 MB/s sim-side CPU/bandwidth waste, no functional impact. Gate behind a settings.json flag in a separate PR.
- The new TCP path has no compression; if a future Mac Docker Desktop regression reintroduces a loopback throughput cap, LZ4 over the same TCP stream is the next lever.


raulmoranguerra:
Pushed follow-up `ee5453f` that gates the sim's UDP LiDAR broadcast off by default (was wasting ~1.5 MB/s of game-thread CPU / loopback bandwidth on packets nobody was consuming). Bridge's UDP-fallback branch sends `enableLidarUdpBroadcast` over the existing RPC client when it activates.

Verified on Windows: with TCP default, `tcpdump 'udp dst port 41500'` captures **zero packets** over a 5s window while `/lidar/Lidar1` continues at 10.5 Hz on the new TCP path.

raulmoranguerra:
**Mac validation 2026-05-14** — found a missing piece, fixed in commit `a39f83b` pushed to this branch.

## Symptom on Mac

`/lidar/Lidar1` silent. Bridge log showed a ~10 Hz reconnect loop:
```
LiDAR stream reconnected (TCP)
LiDAR stream disconnected (payload recv failed)   ← 4 ms later
```

## Root cause

`FSDSRpcServer::Listen()` at line 229 has a comment block explicitly warning that any RPC pushing >2 MB synchronously needs the per-client `SO_SNDBUF` bump that PR #322 removed ("if a future per-client RPC ever pushes >2 MB synchronously, reintroduce SetSendBufferSize"). PR #482 brought TCP-LiDAR (1.5 MB scans) back without honouring that precondition.

The Windows verification didn't catch it because Windows' default winsock send buffer is much larger than macOS's. On Mac, the plugin's first `Send()` filled the small default buffer, the second `Send()` in `SendAll` failed, `StreamLidar` exited, bridge saw FIN.

Diagnosed via a one-line header dump in `lidarStreamThread` — confirmed plugin WAS sending sensible frames (`magic=LIDR`, `frame_id` advancing, `total_points≈95000`), only the 1.52 MB payload was getting cut off.

## Fix in a39f83b

In `FFSDSRpcServer::Listen()` right after `SetNoDelay(true)` on each accepted client, call `SetSendBufferSize(32 MB)` (same value pre-#322 used) and log a warning if the kernel caps below request.

## Verification

After Mac rebuild:
- `/lidar/Lidar1`: **10.021 Hz** (datasheet rate)
- Bridge log: clean frame_id sequence 0, 1, 2, 3, ...
- No payload-recv failures, no reconnect loops

Should now work on both Mac and Windows with the same wire protocol.

raulmoranguerra:
Pulled `a39f83b` and regression-tested on Windows: no behaviour change, no `SO_SNDBUF capped` warning (Windows winsock granted the full 32 MB request), `/lidar/Lidar1` holds **10.0–10.3 Hz** steady over a 25 s sustained-rate check, zero bridge "payload recv failed" / reconnect-loop entries in the log. So the fix is safe to land — Mac gets the partial-send headroom it needs, Windows is unaffected.

PR #486, opened 2026-05-14
feat(mission-control): record bag on session start (#465)

### LINKED ISSUE #465: feat(mission-control): record bag on session start
## Why

Right now I have to ask Claude to start \`ros2 bag record\` inside the dv_pipeline_stack container before each live test. That's friction every time we want diagnostic data, and it puts the responsibility for matching the recording window to the actual session into a side channel. Should be a checkbox on the same UI that starts the session.

## What

Add an optional **Record bag** toggle to EventSetup that, when checked, makes \`/api/event/start\` also kick off an mcap bag-record on the dv_pipeline_stack container; the matching session-stop tears it down. Default: **off** (so it never silently fills disk).

### UI

- Checkbox in \`EventSetup.tsx\` next to the \"Start Session\" button: **\"Record bag (mcap)\"**, unchecked by default. Persist last choice in localStorage so consecutive test sessions don't have to re-tick.
- Disabled while the session is busy (same gating as the other inputs).
- After session starts, show the bag name + path on the active-event panel so you can grab it later without asking.

### Backend

- Extend the \`EventSetup\` POST schema with \`record_bag: bool = False\`.
- In \`main.py:event_start\`, if \`record_bag\`:
  - Compose a name: \`<event_type>_<track>_<YYYYMMDD_HHMMSS>\` (e.g. \`trackdrive_TrainingMap_20260512_153022\`).
  - Spawn \`ros2 bag record -s mcap -a -o /bags/<name>\` inside dv_pipeline_stack via \`docker exec -d\` (we already shell out to docker for sim restarts).
  - Capture the pid + bag path in the in-memory event state.
- In \`/api/event/stop\` (or wherever StopSession lives), if a record pid is tracked, \`pkill -INT\` it inside the container and copy the bag to the host \`bags/\` dir.
- Expose \`bag_name\` and \`bag_state\` (\`recording | stopping | stopped | none\`) on \`/api/event/state\` so the frontend can show it.

### Bag landing path

- Inside container: \`/bags/<name>/\` (already what we use).
- On host: \`bags/<name>/\` next to \`tools/bags/\` — keep them separate from the canonical test-fixture bags under \`tools/bags/\` (which feed \`tools/replay.sh\`).
- \`bags/\` is already in \`.gitignore\` (PR #436).

### Safety

- Default off — no surprise disk usage.
- Free-disk pre-check: refuse to start recording if free space < 10 GiB on the bind-mount. Show a clear error in the UI.
- If \`ros2 bag record\` fails to start, surface a clear toast — session-start should still succeed (recording is opt-in).
- If session-stop happens but the bag-record pid is dead already (e.g. crashed), don't hang on \`pkill\`.

### Out of scope

- No backend changes to bag rotation / size caps — single-file mcap is fine for our laps.
- No multi-topic filtering UI — \`-a\` (all topics) is what we always want.
- No live "currently recording at X MB/s" indicator — too much polling churn for the value. Just a static "recording" badge.

## Acceptance

- [ ] Checkbox visible + functional, defaults off
- [ ] Session start with checkbox ticked → bag file appears in \`bags/<name>/<name>_0.mcap\`
- [ ] Session stop → bag closed cleanly (no truncated final chunk)
- [ ] Free-disk check works (force it to fail in a test)
- [ ] localStorage persists the checkbox state across page reloads
- [ ] Existing test suite (\`tests/test_*.py\`) still passes — new test added for the bag-record flow with a mocked \`docker exec\`

raulmoranguerra:
Closes #465.

## Summary

Replaces the \"ask Claude to docker exec ros2 bag record\" friction with a checkbox in EventSetup. Default off — no silent disk fills. Recording window matches the session window: starts after autonomy is up, stops on Stop Session, bag copies out to host \`bags/\`.

## Backend

- New \`bag_recorder.py\` — pure-Python helpers (compose-name, free-disk pre-check, start/stop via docker exec). Subprocess-injectable for tests.
- \`EventSetup\` schema: optional \`record_bag: bool = False\` (legacy clients unaffected).
- \`event_start\` Phase 3 kicks off recording. Failures don't fail the session — surfaced as \`r.bag.error\` so the operator knows.
- \`pipeline_stop\` tears it down AFTER autonomy teardown (so final on_deactivate messages land in the bag), then \`docker cp\` to host.
- \`/api/event/state\` and the WS telemetry frame now carry \`bag_state\` and \`bag_name\`.
- New env knobs: \`DV_PIPELINE_CONTAINER\`, \`HOST_BAGS_DIR\` — both default to existing conventions.

## Frontend

- \"Record bag (mcap)\" checkbox above Start/Stop, persisted to localStorage.
- Live badges: \`● RECORDING\`, \`◌ starting\`, \`◍ bag saved\`, \`✗ failed\`.
- Start-session response surfaces \`r.bag.name\` / \`r.bag.error\` in the status line.

## Tests

**17 new pytest cases** in \`tools/mission_control/backend/tests/test_bag_recorder.py\`. Docker plumbing mocked via subprocess injection — no docker on the runner required.

Backend: **86/86 pass**.
Frontend: \`tsc -b\` + \`eslint .\` both clean.

## Acceptance (from #465)

- [x] Checkbox visible + functional, defaults off
- [x] Session start with checkbox ticked → recorder spawned
- [x] Session stop → bag closed cleanly via SIGINT (not SIGKILL)
- [x] Free-disk check works (test case force-fails it)
- [x] localStorage persists checkbox state across reloads
- [x] Existing test suite passes — new test for the bag-record flow
- [ ] **Pending live validation**: drive a session with the checkbox ticked, confirm bag lands at \`bags/<name>/<name>_0.mcap\`. Mac + Windows.


raulmoranguerra:
## v2: recorder moved into dv_pipeline_stack (cc340a4)

Initial implementation recorded locally inside `mc_backend`. Smoke tests passed but a full autocross session showed **96% loss on `/lidar/Lidar1`** (0.4 Hz instead of 10), 100% on `/camera/*`, 72% on `/tf`. Root cause: `mc_backend` is forced to `FASTDDS_BUILTIN_TRANSPORTS=UDPv4` (so the StartMission action client works cross-container) and Fast DDS's UDP fragmentation reassembly drops most multi-fragment messages from the SHM-preferring publisher.

**Fix**: move the recorder into `dv_pipeline_stack` where it shares the SHM-tuned DDS context with the publishers. `mc_backend` talks to it via two new `dv_msgs` services hosted by the new `bag_recorder_node` package:

- `/bag_recorder/start` (`dv_msgs/srv/StartBag`)
- `/bag_recorder/stop` (`dv_msgs/srv/StopBag`)

The recorder writes to `/tmp/ifssim_bag_active/` inside the container, then moves the finalised bag to the bind-mounted `/bags/` on stop — same staging+move pattern as `tools/record_bag.sh`.

### Validated live (60s autocross, 1.1 GB bag, 18.7 MB/s sustained)

| Topic | Before | After |
|---|---|---|
| `/lidar/Lidar1` | 0.4 Hz | **10.0 Hz** ✅ |
| `/lidar/Lidar1/viz` | 1.4 Hz | 10.0 Hz |
| `/tf` | 29 Hz | 109.7 Hz |
| `/imu` | 320 Hz | 329.7 Hz |
| `/control_command` | skipped¹ | 38.7 Hz |

¹ fs_msgs/dv_msgs weren't on `AMENT_PREFIX_PATH` in `mc_backend`; rosbag2 logged "unknown type" warnings and skipped the topic. Sourcing `/dv_pipeline_stack_ws/install/setup.bash` in `bag_recorder_node` fixes that.

### Tests
- `pipeline/bag_recorder_node/test/test_recorder.py` (13 tests) — subprocess + staging + SIGINT
- `tools/mission_control/backend/tests/test_bag_recorder.py` (13 tests) — rclpy service-client wrapper

PR #487, opened 2026-05-14
feat(slam): KISS-ICP pose anchor — Pattern A cascade fix

### LINKED ISSUE #485: feat(cone_slam): KISS-ICP pose source as cascade fix (Pattern A)
## Goal

Adopt KISS-ICP as a pose source feeding the existing cone_slam factor graph, fixing the structural cascade documented in `docs/archive/2026-05-06_slam-phase2-cascade-frozen.md` without ripping out the cone-landmark identity layer that `path_planning` consumes.

## Background

Cone-only data association cannot survive > 1 m of predicted-pose drift in cone-poor scans. Seven months of incremental mitigations (proximity veto, spike detector, sanity check, velocity prior) have pushed cascade onset from t≈52 s to t≈63 s but not closed it. The root cause is structural: when 0–1 cones are in FoV, the factor graph has no global-frame anchor and iSAM2 finds a cheaper minimum by rotating the world.

GLIM was attempted (PR #3) and failed live (76% drift, sustained-load crashes on the CPU module). KISS-ICP is simpler, lighter, and addresses the same root cause without GLIM's dependency / memory baggage.

## Pattern A — drop-in pose source

1. Run `kiss_icp` as a sibling ROS 2 node consuming `/fsds/lidar/Lidar1`, publishing `/odom_lidar`.
2. Inside `cone_graph_slam_node`, when `/odom_lidar` is fresh, stage a tight `PriorFactorPose3` on `X(k)` (σ ≈ 5 cm / 10 mrad).
3. Cascade goes away because global rotation is pinned every scan; cone bearing-range factors become pure landmark refinement.
4. All existing cone_slam plumbing (DA, landmark_db, /Conos persistent IDs, replay harness) preserved.

Env-gated via `IFSSIM_USE_KISS_ICP_PRIOR=1` so cone_slam can run in either mode.

## Acceptance criteria

- [ ] KISS-ICP node publishes `/odom_lidar` at sim LiDAR rate (10 Hz)
- [ ] Replay against existing bag (`tools/bags/lap_attempt_*`): `/odom_lidar` vs `/testing_only/odom` drift < 0.5 m at lap end
- [ ] With prior factor enabled: full lap with NO cascade in `replay_slam.py` regression
- [ ] Sensor rates unchanged (`/Conos`, `/imu`, `/motor_rpm` all baseline)
- [ ] Tests: new unit test for the prior-factor staging path

## Out of scope (follow-ups)

- Pattern B refactor (delete cone_slam pose estimation entirely, keep landmark-only graph)
- Loop closure layer for endurance laps
- GPS factor (PR C from cone_slam handoff) — superseded if KISS-ICP works

raulmoranguerra:
## Summary

Adds KISS-ICP scan-to-map odometry as a sibling lifecycle node and wires its output into cone_slam as a per-scan `PriorFactorPose3` on `X(k)`. Off by default; flip `IFSSIM_USE_KISS_ICP_PRIOR=1` to enable.

Closes #485 (Pattern A — the cascade fix recommended in §9 of the SLAM audit on dev). Cone-only DA cascade documented in `docs/archive/2026-05-06_slam-phase2-cascade-frozen.md` is structural — iSAM2 has no global-rotation anchor in cone-poor scans. KISS-ICP provides that anchor, dense geometry every scan.

GLIM was the original swap target (`docs/history/glim_integration.md`); it failed live in PR #3 (76% drift + sustained-load crashes on the CPU module). KISS-ICP is simpler (one pip dep, no submap memory pressure, no GTSAM bump) and addresses the same root cause.

## Architecture

```
                  ┌──── kiss_icp_node ────┐
/lidar/Lidar1 ────┤  KISS-ICP @ 10 Hz     ├──── /odom_lidar (Odometry)
                  └───────────────────────┘                │
                                                           ▼
                                        ┌─── slam_node (cone_slam) ───┐
/imu, /Conos_raw, /motor_rpm  ──────────┤  PriorFactorPose3 on X(k)   ├── /Conos
                                        └─────────────────────────────┘
```

* Pattern A: `X(k)` stays a variable in cone_slam's factor graph; the KISS-ICP pose becomes a tight prior every scan.
* Pattern B (cone_slam pose-elimination refactor) is a deliberate follow-up after Pattern A demonstrates the cascade fix.

## What's in this PR

* **`pipeline/kiss_icp_wrapper/`** — new ament_python package. LifecycleNode wrapping `kiss_icp.kiss_icp.KissICP`. Subscribes `/lidar/Lidar1` BEST_EFFORT, parses PointCloud2 the same format-agnostic way `cone_detection_node` does, calls `register_frame` per scan, publishes `/odom_lidar` (nav_msgs/Odometry, `header.frame_id="kiss_odom"`, `child_frame_id="base_link"` — `base_link↔fsds/Lidar` is identity per `ifssim_ros_wrapper.cpp:1370` so no static-TF composition needed). FS-tuned config: 50 m max range, 0.5 m voxel.
* **`Dockerfile`** — `pip install kiss-icp`. Pure Python, no native compile. Resolver lands cleanly on top of the existing numpy==1.24 pin; numpy re-pinned defensively after.
* **`pipeline.launch.py`** — `kiss_icp_node` added as autonomy lifecycle node, before `slam_node` in bring-up order.
* **`mode_manager.AUTONOMY_LIFECYCLE_NODES`** — `kiss_icp_node` inserted between cone_detection and slam_node.
* **`docker-compose.yml`** — bind-mount the new package + propagate `IFSSIM_USE_KISS_ICP_PRIOR` env var.
* **`factor_graph.FactorGraph.stage_kiss_icp_prior()`** — new method. `PriorFactorPose3` on `X(self._k)` with sigmas tuned for KISS-ICP-on-FS-track: σ_xy=5cm, σ_z=20cm, σ_yaw=0.02 rad, σ_roll_pitch=0.10 rad.
* **`cone_graph_slam_node`** — `_sub_kiss` BEST_EFFORT subscription gated by env. `_kiss_init_pose` snapshotted at `_finish_calibration` time (mirrors the `_gt_init_pose` pattern). Per scan in `_on_cones`: if flag on + anchor captured + sample < 0.5 s old, compose `T_slam = T_kiss_anchor⁻¹ · T_kiss_now` and pass to `stage_kiss_icp_prior`.

## Test plan

- [x] Image builds clean (`pip install kiss-icp` resolves on top of numpy==1.24)
- [x] All nodes spawn at launch; `ros2 node list` shows `/kiss_icp_node` healthy
- [x] `kiss_icp_node` lifecycle: `unconfigured → inactive → active` transitions succeed
- [x] With `IFSSIM_USE_KISS_ICP_PRIOR=0` (default): `/odom_lidar` has 1 publisher / 0 subscribers — slam_node ignores it. Pre-#485 path bit-identical.
- [x] With `IFSSIM_USE_KISS_ICP_PRIOR=1`: `/odom_lidar` has 1 publisher / 1 subscriber. `ros2 node info /slam_node` shows `/odom_lidar` in subscribers.
- [ ] **Live or replay**: full-lap test against an existing bag with the flag on; verify cascade gone (currently hits at t≈63 s on `lap_attempt_*`). Next session.
- [ ] Sensor rates unchanged baseline (`/Conos`, `/imu`, `/motor_rpm`). Live-validate next session.

## What's NOT in this PR

* Replay validation against an existing bag — pending the next session.
* Pattern B refactor (X(k) as input not variable) — separate follow-up.
* Unit test for `stage_kiss_icp_prior` — pending; smoke-tested via lifecycle bring-up.
* TF composition for non-identity lidar mounts — currently identity; if the lidar ever moves off `base_link` origin, see TODO in `kiss_icp_node.py`.

## Refs

* Issue: #485
* SLAM audit: chat-only, summarized in #485 description
* Cascade post-mortem: `docs/archive/2026-05-06_slam-phase2-cascade-frozen.md`
* GLIM live failure (prior swap attempt): `docs/history/glim_integration.md` §15


raulmoranguerra:
## Live-test results — Pattern A is **blocked** by KISS-ICP tracking failure

### Test setup
- Sim: IFSSIM v0.1.0 Windows build, sim active, trackdrive mission activated
- Recording: bags/kiss_pattern_a_v1/ (635 MB MCAP, ~22 s actual drive)
- Container: `IFSSIM_USE_KISS_ICP_PRIOR=1`, both kiss_icp_node and slam_node active

### Bug found + fixed (e207f98)
First run crashed every scan with `'NoneType' object has no attribute 'ravel'`. Second attempt (after passing `np.zeros(N)` for timestamps) SIGABRT'd the whole node mid-drive via Sophus' `SO3::exp` NaN ENSURE. **Only** `np.array([])` works as the "no per-point timestamps" sentinel — matches the upstream KITTI reader pattern. Now ships in this PR; live verified at 50 scans / 5 s, 0 failures, ~13 ms/scan.

### Cascade outcome — much worse than baseline
| | Baseline (cone-only) | Pattern A (KISS-ICP prior) |
|---|---|---|
| Cascade onset | ~63 s | **t = 4 s** |
| Final SLAM-vs-GT error | ~30 m | **555 m** |
| GT distance driven | — | 25 m |

### Root cause — KISS-ICP isn't tracking motion on the sim LiDAR cloud
Body-frame comparison over the 21.7 s drive window:

| | GT (body frame) | KISS-ICP |
|---|---|---|
| Final position | (+24.07, +6.41) m | **(-0.99, +0.00) m** |
| Total displacement | ~25 m | ~1 m |
| Final yaw | +90° → +128° (turned 38°) | 0° → 0° (**no rotation**) |

KISS-ICP locked into a wrong local minimum at -1 m offset on the second scan and stayed there for the entire drive. It then fed that lock to the iSAM2 prior at tight σ (5 cm / 0.02 rad), which dragged cone_slam off course in seconds — far faster + larger than the cone-only baseline ever did.

### Read on the failure mode
Diag throughput was fine: 39-52 scans / 5 s, 0 failures, 12-50 ms/scan. KISS-ICP was processing every cloud — it just registered them all to ~identity. Most likely explanations (no fix proposed yet, just hypotheses):

1. **FS-track scenes are too symmetric for KISS-ICP's defaults** — large flat ground + ~20 cones with low height variance. ICP's adaptive threshold may be collapsing onto the ground plane (which is rotation/translation-invariant within itself).
2. **Initial threshold too generous** (2 m default) — with the car barely moved scan-to-scan, the constant-velocity initial guess at identity is close enough that ICP converges to identity rather than the true motion.
3. **Voxelization eats motion signal** — at 0.5 m voxel, single-step car displacement at low speed is sub-voxel.

### Decision

**Do not merge** PR #487 in its current form. The integration is correct and the wrapper is healthy, but the upstream registration quality on FS-track LiDAR is a non-starter. Three follow-up paths, ranked:

1. **Diagnose KISS-ICP empirically** — record `/lidar/Lidar1` raw + `/testing_only/odom` to a bag and run `kiss_icp_pipeline` against it offline as a standalone tool. If standalone KISS-ICP also fails, the algorithm doesn't fit this LiDAR pattern. If it works, our wrapper has a bug.
2. **Tune KISS-ICP for FS tracks** — smaller voxel (0.2 m), tighter `initial_threshold` (0.3 m), different `max_range`. Maybe filter ground via `min_range_z`.
3. **Pivot to GPS-factor PR C** — the original §9 fallback in the audit. ~30 LOC, /gps already published, doesn't depend on dense registration working.

The empty-timestamps fix (e207f98) is a real bug fix and is independently mergeable. The rest of the PR stays as work-in-progress until KISS-ICP tracks.

Bag preserved at `bags/kiss_pattern_a_v1/` (635 MB MCAP, replay-able with the `IFSSIM_USE_KISS_ICP_PRIOR=1` container) for follow-up triage.

raulmoranguerra:
Closing — Pattern A swap is the wrong fix for this stack.

Live-test summary on PR head e207f98:
- KISS-ICP plumbing works (lifecycle, env-gate, anchor, prior staging) ✓
- KISS-ICP itself does NOT track on the FS-track LiDAR scene: drives 24 m / 38° → reports -1 m / 0° rotation, locked at near-identity from scan 2 onwards
- 16-combo config sweep (voxel × init_threshold × min_range) found NO combination that tracks
- Standalone repro outside the wrapper confirms it's algorithm × scene, not a wrapper bug
- Diagnostic: most LIO algorithms (FAST-LIO2, LIO-SAM, Point-LIO, DLIO, GLIM) all need per-point timestamps for deskewing, which the FSDS sim does NOT emit. Swapping to any of them would require a sim-side change to `FSDSLidarSensor` first.

Preserved for future reference:
- The empty-array timestamps sentinel fix (commit e207f98) is a real bug that would bite anyone using kiss-icp 1.3.0 from Python. Documented in the commit body.
- The cone_slam consumer pattern (PriorFactorPose3 on X(k) gated by env var, anchor snapshotted at calibration end) is producer-agnostic — reusable when a working pose source exists.

Going back to dev. Issue #485 stays open as "Pattern A approach rejected; needs a working LiDAR pose source first; sim probably needs per-point timestamps before any LIO will track."

Issue #497, opened 2026-05-14
fix(docker): keep live ROS edits without Windows source bind mounts


sjrom47:
## Observation

Windows Docker Desktop runs slowly when ROS Python source trees are bind-mounted from the host into `dv_pipeline_stack`. Cold autonomy startup can spend tens of seconds in imports, package discovery, and Numba cache checks, and developers still want live Python edits without paying that bind-mount cost.

## Diagnosis

#490 removed the expensive source bind mounts and added `.dockerignore`, which made rebuilds much cheaper. The remaining gap is live iteration: image rebuilds are reliable, but slower than the old edit/restart loop for Python-only ROS package changes. Docker named volumes stay on Docker Desktop's Linux-backed storage, and Compose Watch can sync host edits into those volumes without mounting the entire package tree through the Windows filesystem.

## Fix

Use Docker-managed named volumes for the runtime ROS source directories in `dv_pipeline_stack`, and add `develop.watch` sync rules from the host package directories into those volumes. Keep the image-time `colcon build --symlink-install` baseline so Python files already present at build time resolve through `/dv_pipeline_stack_ws/src`, while documenting when a rebuild is still required.

Keep #490's other improvements: DDS profile copied into the image, `.dockerignore`-reduced build context, and `ifssim_bags` as a named volume for fast StopBag finalization.

## Scope

- Compose-only follow-up for `dv_pipeline_stack` runtime mounts and watch rules.
- No autonomy behavior change intended.
- Python source edits should sync live; C++, `.msg`, and `setup.py` changes still require rebuild/re-link.

## Acceptance

- `docker compose config --quiet` passes.
- `dv_pipeline_stack` no longer bind-mounts ROS source package trees from the host.
- `docker compose watch dv_pipeline_stack` syncs edits for `pipeline/*` and `ros2/src/*` packages into named volumes.
- The bag landing directory continues to use `ifssim_bags`, with host retrieval through the existing bag helper scripts.

Issue #496, opened 2026-05-14
Custom EKF-SLAM with cone landmarks — recommended SLAM rewrite path (LIMOncello ruled out in #504)


raulmoranguerra:
Saved as the backup plan in case LIMOncello (currently being smoke-tested in feat/limoncello-smoke) doesn't track FS-DV scenes cleanly. If LIMOncello works, this issue stays parked; if it fails the smoke test, this is the next path.

## Premise

Single-frame Kalman SLAM, state vector grows with discovered landmarks. **All math in `numpy`, no GTSAM, no scipy.optimize, no factor graphs.** Every line auditable. ~850 LOC total + ~400 LOC of tests.

Drop-in replacement for `cone_graph_slam_node` — same topics, same TF chain, same downstream consumers (path_planning, mission_control). Switch via launch flag, keep cone_slam alongside until parity is proven on `lap_postveto_20260511_183638`.

## State vector

```
x = [ pose_x, pose_y, theta,           # vehicle pose (world frame)
      v_x, v_y, omega,                  # body-frame velocity + yaw rate
      b_gz, b_ax, b_ay,                 # IMU biases (gyro_z, accel_x, accel_y)
      L1_x, L1_y, L2_x, L2_y, ... ]    # landmark positions (world frame)
```

Fixed 9-element vehicle slice, then landmarks appended. Covariance grows accordingly. 200 cones → 409×409 matrix, trivial in numpy. Landmark metadata (colour, last_seen_step, observation count, confidence) lives in a parallel Python dict — decouples bookkeeping from optimization.

## Predict step (Coriolis terms in from day one)

Discrete-time motion model at IMU rate (~400 Hz):

```
theta'  = theta + (omega - b_gz) · dt
v_x'    = v_x + (a_x_body - b_ax) · dt + omega · v_y · dt          # ← Coriolis
v_y'    = v_y + (a_y_body - b_ay) · dt - omega · v_x · dt          # ← Coriolis
pose_x' = pose_x + (v_x · cos(theta) - v_y · sin(theta)) · dt
pose_y' = pose_y + (v_x · sin(theta) + v_y · cos(theta)) · dt
```

The Coriolis cross-product (the bug `OdometryFilter` has — see #447 audit) is in the EKF predict from day one, so SLAM pose stays clean even if `/odom` (from `odometry_filter_node`) keeps drifting. Biases evolve as random walk. Landmarks are static.

[...]

PR #504, opened 2026-05-15
feat(slam): LIMOncello smoke test — closed unmerged, documented dead end

raulmoranguerra:
**This PR is not going to merge.** It exists to preserve the smoke-test findings as a permanently-browsable record so future engineers don't reattempt LIMOncello on FS-DV scenes without the prior data.

## TL;DR

LIMOncello (Iterated Error-State Kalman Filter on the SGal(3) manifold for LiDAR-Inertial Odometry, https://arxiv.org/abs/2512.19567) **does not track FS-DV scenes** cleanly. The failure mode is algorithmic (LIO scan-matching can't anchor pose on flat-ground + sparse-vertical-cones scenes), not a configuration issue we missed.

Matches the prior memory note about FAST-LIO2 / fast_LIMO / GLIM all failing on the same scene — same family of algorithms, same family of failure.

## What was tried, in order

1. **Vendored LIMOncello via fork** (`isc-fs/LIMOncello:ifssim-main`) with Livox driver deps gated behind a CMake option so we don't need three unobtainable Livox apt packages
2. **Bridge: per-point FLOAT64 `timestamp` field** on `/lidar/Lidar1` (LIMOncello's HESAI handler requires it)
3. **Bridge: `/lidar/Lidar1_rep103` companion topic** with Y negated so LIMOncello sees REP-103 axes consistent with `/imu`
4. **`SensorDataQoS` on the fork's IMU + LiDAR subscriptions** — replaces stock RELIABLE which silently mismatched our bridge's BEST_EFFORT publishers
5. **Bag rewriter** (`tools/limoncello/upgrade-bag-lidar-timestamp.py`) — synthesizes both `/lidar/Lidar1` (post-#497 timestamp layout) and `/lidar/Lidar1_rep103` from pre-#497 bags
6. **SE(2)-Umeyama RMSE analyzer** (`tools/limoncello/compare-pose-rmse.py`) — aligns LIM trajectory against `/testing_only/odom` GT before computing RMSE so frame-rotation-only error doesn't inflate the score
7. **Lichtblick `limoncello_eval.json` layout** — 3D + xy + velocity + yaw-rate side-by-side
8. **Calibration tuning** — `time: 0.0` (skip entirely) gave best results; `gravity_align: true` regressed; `time: 3.0` with bias estimation polluted the bias because autocross bags start moving at ~2.2 s

## Final smoke test results

Bag: `autocross_track_20260404_013721_20260514_145834` (60 s, 603 LiDAR scans @ 10 Hz, 19,882 IMU @ 329 Hz, full sensor coverage)

| metric | value |
|---|---|
| RMSE (xy, SE(2)-aligned) | 34 m |
| Max xy error | 53 m |
| Final-frame error | 12 m |

Time-series breakdown (the smoking gun):

| t | GT v_body (forward, lateral) | LIM v_body |
|---|---|---|
| 0.1 s — stationary | (+0.00, −0.00) | (−0.00, +0.00) ✓ |
| 3.0 s — first motion | (+1.29, +0.00) | (+1.88, +0.38) ✓ same direction, 1.5× magnitude |
| 6.0 s — cornering starts | (+2.74, +0.05) | (+4.32, +0.06) ✓ still forward |
| **10.0 s — CASCADE EVENT** | (+2.90, −0.01) | (−12.56, +4.60) ❌ |
| 15.0 s — post-cascade | (+2.92, −0.06) | (−26.09, −17.21) ❌ |

LIMOncello tracks correctly for ~6 seconds, then loses scan-matching lock during the first corner and the filter cascades. Yaw rate matches the IMU perfectly per-tick (correct integration), but yaw drifts unbounded because there isn't enough geometric structure in the scene to constrain it.

## Why this is structural, not config

Both bag-replay (`lap_postveto_20260511_183638`, 117 s) and live-test runs hit the same cascade. The convention bug hypothesis was explicitly tested:

- `/imu` accel reads +9.86 m/s² in Z at rest → REP-103 standard (Z up). ✓
- `/lidar/Lidar1_rep103` is verified byte-for-byte (X/Z/intensity preserved, Y negated) — see commit b600737's diagnostic output. ✓
- `gravity_align: true` to estimate any residual IMU tilt regressed results (RMSE 34 → 128 m).
- Body-frame velocities match GT for the first 6 s of motion (no convention bug); only after the cascade do signs flip.

The conclusion: this is feature-poverty of the FS-DV scene defeating scan matching, not a setup problem.

## What ships forward despite this PR not merging

The IFSSIM-side commits in this branch (bridge per-point timestamp, `/lidar/Lidar1_rep103` companion, bag rewriter, RMSE analyzer, Lichtblick layout) are all LIMOncello-specific or LIMOncello-shaped — they aren't useful without a LIO consumer. They will be re-introduced from scratch if and when a viable LIO candidate appears.

The fork `isc-fs/LIMOncello` stays alive — it carries the Livox-optional patch (submittable upstream as a quality-of-life PR even though we're not using it).

## Recommended next path

#496 — custom EKF-SLAM with cone landmarks. Cones provide unambiguous discrete landmarks (vs. point cloud blur scan-matching keeps failing on), which sidesteps the yaw-constraint problem this PR exposed.


raulmoranguerra:
Closing unmerged per the verdict in the PR body — LIMOncello cannot anchor pose on FS-DV's flat-ground + sparse-vertical-cones scenes regardless of axis convention, extrinsics, or calibration tuning. Pivoting to #496 (custom EKF-SLAM with cone landmarks). PR + commits preserved here as a permanent record so this isn't reattempted blind.

Issue #386, opened 2026-05-10
Tune gyro bias estimation — /odom yaw drifts +0.9°/s during motion


raulmoranguerra:
**Surfaced by the lap drift analysis in PR #378.**

`/odom` yaw drifted **+37° over a 41 s motion window** — effective gyro_z bias of ~0.9°/s. BMI088 typical post-stationary-calibration bias is <0.5°/s. We're 2× spec.

### 1. Longer stationary calibration window

`pipeline/sim_supervisor/sim_supervisor/odometry.py` `CALIBRATION_SECONDS = 3.0`. Bumping to 5.0 s gives ~67 % more samples (1200 → 2000 at 400 Hz) → tighter bias mean estimate. Cheap.

### 2. Pre-calibration motion-sanity check

The car's spawn-time settling motion may bleed into the bias estimate (suspension settles for ~200–300 ms after spawn, IMU sees the residual accel). Add a pre-window check: if max(|accel - g|) > threshold over the first 0.5 s, restart the calibration window with a 1 s settling delay.

### 3. Mid-drive bias re-estimation

Detect long straights (`yaw_rate < 0.05 rad/s` sustained for > 3 s) and slowly pull `gyro_bias` toward the implied value:

```
gyro_bias_new = (1 - λ) * gyro_bias_old + λ * mean(gyro_z over straight window)
```

with `λ ≈ 0.01` per straight detected. SLAM does this implicitly via iSAM2's bias factors; `/odom` could implement a simpler version standalone.

## Acceptance

Post-tuning, `/odom` yaw drift over a 60 s drive < 10° (against `/testing_only/odom` GT yaw, motion-window-anchored).

## Related

- #380 (RPM_TO_MS calibration) — independent but should land first; the velocity error currently dwarfs the yaw error in `/odom`-vs-GT residuals
- #385 (Q2 IMU rate review) — longer calibration window may interact with rate choice

raulmoranguerra:
**Update 2026-05-12** — issue title is partially mis-framed; clarifying.

The 9-state EKF in PR #451 (draft) estimates [b_ax, b_ay, b_gz] as filter states. Today's in-loop test against the complementary filter result: **tied at 64s autonomy.**

[...]

PR #550, opened 2026-05-23
fix(odometry_filter): freeze gyro bias + make steering correction diagnostics-only

raulmoranguerra:
## Summary
Two structural EKF fixes uncovered by the #386/#447 bag-driven investigation. **They make /odom's published yaw rate correct on the signal level** (no more 1-2 °/s phantom bias during corners, no more ±12 °/s spikes from steering correction polluting OMEGA). They do **NOT** fix the recent short-bag spin-outs — those happen in the first corner inside 10 s and are a control-loop issue separate from /odom.

## #1 — Bias freeze: `K[BG_Z] = 0` in all three correction functions

`correct_rpm`, `correct_nhc`, and `correct_steering` all have H rows that don't include BG_Z. But the standard EKF update produces a non-zero `K[BG_Z]` through `P[BG_Z, OMEGA]`, which is non-zero because `F[OMEGA, BG_Z] = -1` in the propagation step. So any non-gyro observation pulls the gyro-bias state via cross-correlation.

On bag `_113508` (analysed in #447 step A) this manifested as `bg_z` walking ±1.8 °/s during a 15 s sustained corner. The mechanism: kinematic-bicycle steering observations carry MODEL ERROR (tire slip, suspension flex), not gyro bias information, but the EKF distributes the OMEGA-residual between OMEGA and BG_Z proportional to the cross-covariance.

**`sigma_bg_walk` tuning can't fix this** — it controls PROCESS noise, not observation gain through cross-correlation. We tried 1e-4 → 1e-5 in a separate branch (`fix/ekf-bg-walk-tighten`, not pushed); the tighter value let the bias still walk to -3.9 °/s in a single second on bag `_223932`.

The principled fix is to set `K[BG_Z] = 0` in each non-gyro correction. Bias still propagates via Q and can still be re-estimated by stationary calibration. Accel-bias updates (`K[BA_X]`, `K[BA_Y]`) in correct_rpm are intact.

**Validated on bag `_225200`**: `b_est` stays at exactly +0.0084 °/s (the calibration value) for the entire run, vs the ±1.8 °/s walk seen with the previous build.

## #2 — Steering correction: diagnostics-only

After bias-freeze, `_225200` still showed sporadic ±12 °/s spikes in `/odom.angular.z` vs raw gyro. Root cause: `correct_steering` still updates `x_(OMEGA)` via `K[OMEGA] · (omega_pred - x_(OMEGA))` when the residual stays just under the slip-gate threshold (0.3 rad/s = 17 °/s — too loose for transient FS-DV corners). Between IMU callbacks (which reset OMEGA to `gyro − bias`), the 100 Hz publish timer can catch the steering-corrupted state and that polluted value integrates into θ.

The IMU is clean (raw `gyro_z` offset vs GT measured at 0.002 °/s on straights in `_113508`) and bias is now frozen, so `x_(OMEGA) = gyro − bg_calibrated` IS truth — no kinematic anchor needed. `correct_steering` still computes `diag_.yaw_residual` + `diag_.slip_flag` (the NHC step depends on `slip_flag`), but early-returns before the EKF update.

If real-car runs eventually show gyro drift, revisit with a much tighter slip threshold (~0.05 rad/s) or a model that doesn't shed model-error onto ω. The disabled block is kept in source as "unreachable but documented" so re-enabling is a one-line revert.

## Test plan
- [x] Bag-replay `_225200` shows `b_est` constant at +0.0084 °/s through the whole run
- [x] Bag-replay `_225756` shows mean |odom − gt| under 0.5 °/s and no ±12 °/s spikes
- [ ] Real-car run — verify gyro doesn't drift over a 60+ s autocross. If it does, this PR's note about a tighter `slip_threshold` applies.

## Relates to
- #386 (gyro bias / odom yaw drift)
- #447 (OdometryFilter position drift)


raulmoranguerra:
Closing — the bias-freeze patch has a covariance-symmetry bug.

Setting \`K(BG_Z) = 0\` AFTER computing \`K = P · Hᵀ / S\` makes the covariance update \`P_new = (I − K·H) · P\` asymmetric. Specifically:
- BG_Z row of P_new stays unchanged
- BG_Z column of P_new still decreases via \`K[i] · P[OMEGA, BG_Z]\` for i ≠ BG_Z

Over many ticks P drifts non-symmetric → eventually non-PSD → garbage covariance values published in \`/odom.*.covariance\`. Downstream consumers (path_planning via TF chain) misbehave: bag analysis shows \`/Path\` publish rate dropping from 10 Hz → 0.6 Hz with this patch applied, planner \`_on_cones\` callback stops firing entirely.

Correlation across yesterday's bags:
| Bag | EKF state | /Path rate |
|---|---|---|
| _113508 | dev (no patch) | 10.0/s |
| _222555 | sigma=1e-5, didn't apply | 9.6/s |
| _223932 | sigma=1e-5, applied | 0.6/s — stops |
| _225200 | bias-freeze K[BG_Z]=0 | 6.8/s, gaps |
| _225756 | bias-freeze + steering disabled | 0.6/s — stops |

The original /odom-yaw-drift symptom is real (bag _113508 showed +0.5 °/s bias on straights, +1.8 °/s plateau on long corners). But the naive K[i]=0 fix isn't safe. A proper Schmidt-Kalman partitioned update or Joseph-form covariance update is needed if we want to lock the bias state against model-error misattribution. Re-attacking in a separate PR.

PR #552, opened 2026-05-24
fix(odometry_filter): freeze gyro bias from non-gyro corrections (Joseph form)

raulmoranguerra:
## Summary
Same goal as #550 (closed): the 9-state EKF mis-attributes kinematic-bicycle / RPM / NHC residuals to gyro bias via the F-matrix coupling, causing `bg_z` to walk ±1.8 °/s during sustained corners (bag `_113508`, #447 step A). #550 fixed the bias update but broke the covariance — this PR fixes both.

## What was wrong with #550
#550 set `K(BG_Z) = 0` after computing the standard Kalman gain, then applied `P_new = (I − K·H) · P`. That update form is only symmetric when K is the *exact* optimal Kalman gain; with a hand-zeroed component, P drifts asymmetric over many ticks. Eventually P becomes non-PSD, the `/odom.*.covariance` fields go garbage, and downstream consumers via the TF chain misbehave. **Empirically: `/Path` publish rate collapsed from 10 Hz → 0.6 Hz on the bag, planner `_on_cones` callback stopped firing entirely.** That's the bug the closing postmortem on #550 describes.

## What this PR changes
Three correction functions in `pipeline/odometry_filter/src/odometry_filter.cpp` (`correct_rpm`, `correct_nhc`, `correct_steering`):

1. **`K(BG_Z) = 0`** before applying the state update — the actual bias freeze.
2. **Joseph-form covariance update**:  
   `P_new = (I − K·H) · P · (I − K·H)ᵀ + K · R · Kᵀ`  
   Preserves symmetry and PSD for **any** K, optimal or modified. This is the canonical recipe when you use a non-optimal gain (e.g. Schmidt-Kalman "considered state", joint-gain partial updates, etc).

## Validation
Bag `_20260524_123201` (60 s autocross drive):

| Metric | dev | this PR |
|---|---|---|
| `/odom` yaw-rate offset on straight (61.6-64.8s) | +0.526 °/s | **+0.046 °/s** (11× reduction) |
| `/Path` publish rate | 9.0 Hz | **9.8 Hz** (no covariance corruption) |
| `b_est` on straight | ~+0.5 (was walking) | **~+0.02 (= IMU's zero-rate residual)** |

So:
- Bias is genuinely frozen at calibration value
- `/odom` yaw rate now matches raw IMU within 0.05 °/s  
- Joseph form preserves covariance properties → path_planning isn't choked

## What this does NOT do
This doesn't make the lap complete. The lap-blocker is **SLAM cone-DA cascade at the late-corner geometry**, not `/odom` yaw drift (today's bag still fails in the same region with the fix applied). The fix is worth landing on its own merits — clean signal, no harm, removes a 0.5 °/s baked-in error from the filter output.

## Test plan
- [x] Bag-replay analyzer confirms `/odom` yaw-rate offset reduced 11× on bag `_20260524_123201`
- [x] `/Path` publishes at full rate (no covariance-collapse failure mode)
- [x] Joseph form vs standard form: built and ran live, container source verified via `tools/refresh-bridge.sh` canary check (now uses #548 volume-drop guard)

## Relates to
- Replaces closed #550
- #386 (gyro bias drift)
- #447 (OdometryFilter position drift)


raulmoranguerra:
Superseded by [#NEW](will-be-replaced) — same bias-freeze + Joseph form, plus NHC slip-aware sigma and per-correction Schmidt-Kalman partitioning. Validated on bag `_211619`: /odom drift dropped from 56 m → 1.4 m (40× reduction), integrated yaw error +60° → +1.3° (45× reduction), SLAM cone-DA cascade gone.

raulmoranguerra:
Replaced by #555.

PR #562, opened 2026-06-11
fix(tools): expand refresh-bridge canary list (post-migration paths)

raulmoranguerra:
## Summary
Recovers the canary-list expansion from the pre-submodule-migration v0.1.3 stash and adapts it to the new layout. The stale-volume verification in `tools/refresh-bridge.sh` now covers **15 files instead of 3**, so edits in `control`, `path_planning`, `cone_detection`, `mode_manager`, `factor_graph`, and the `odometry_filter` C++ get the same host-vs-container sha256 check that previously only guarded the EKF header + two cone_slam files.

## Post-migration adaptation
- `sim_supervisor` moved out of the pipeline submodule to `ros2/src/`, so its canary path is `ros2/src/sim_supervisor/...`.
- The host→container path mapping now strips either a `pipeline/` (submodule) or `ros2/src/` prefix — both collapse to `/dv_pipeline_stack_ws/src/<pkg>/...` in the container.

## Test plan
- [x] `bash -n` clean
- [x] all 15 canary host paths resolve in the working tree (submodule checked out)
- [ ] full `tools/refresh-bridge.sh` run confirms the expanded check passes against a freshly-recreated container


raulmoranguerra:
Closing during a 2026-09 cleanup pass — the defensive canary check this adds is real value on paper, but Docker Compose's develop.watch has been reliable since #490 and no one has hit the drift this would have caught in the intervening months. Reopening later if we ever see a source-out-of-sync bug in the wild is trivial. Not a rejection of the code, just a call that the maintenance surface isn't worth it right now.

PR #566, opened 2026-07-10
chore(pipeline): bump submodule feat/8 (stale) → v1.0.0

raulmoranguerra:
## Summary
Bumps the `pipeline/` submodule pin from `feat/8-dvpc-race-deploy` (2026-07-03, stale) to the tagged release **`v1.0.0`** (`a9b6bbc`, 2026-07-10).

## Why
The pin sat on `feat/8`, branched one day before the **2026-07-04 IMU-topic standardisation**. `feat/8` still carried `REMAP_IMU_CAR = /imu→/imu/data_raw`, which contradicts the real uDV firmware (publishes `/imu`) and the current pipeline (`dev`/`v1.0.0` dropped the remap — nodes subscribe `/imu` directly). That stale remap is what made the sim↔car bag-lift analysis (#564) surface a phantom "firmware gap."

## Safe forward move — nothing lost
- `v1.0.0` is a **strict superset of `feat/8`**: `git log v1.0.0..feat/8` is empty (feat/8's race-deploy work is already merged into the v1.0.0 line).
- `v1.0.0` also contains all of `dev` (`dev...v1.0.0` = 0 behind / 9 ahead) + the `/imu` contract fix + the release.
- Pins to a **tag** (reproducible) rather than a moving branch.

## Relationship to #564
[#564](https://github.com/isc-fs/IFSSIM/pull/564) (sim↔uDV topic parity) was analysed against the stale feat/8; its `lift_to_car.sh` was already corrected to match the real `/imu`-direct contract. This bump makes the sim's pipeline actually *be* on that contract, so a car-parity bag test is faithful. Land either order; both stand alone.

## Test plan
- [x] `v1.0.0` strictly contains `feat/8` (no commits dropped)
- [ ] CI: Bridge colcon build green against the v1.0.0 submodule
- [ ] `tools/refresh-bridge.sh` full rebuild + healthy container on v1.0.0


raulmoranguerra:
Superseded — closing to avoid regressing the submodule pin.

This PR bumps `pipeline` to **v1.0.0**, but the pin has since moved well past that:

- `960c3d0` bumped it to **v1.0.1** (which contains the #41 Stanley limit-cycle de-tune and the #36 cone-detection latency work)
- `aa0cdaa` (on `feat/bag-parity-on-501`) moves it onto **dev + 2**, picking up `max_steer_deg` 28 → 18.2 (#59), `deadband_v` 0.2 → 0.05, `throttle_max` 0.6 → 0.2, the `/slam/final_lap` stop-anchor gating, the signed throttle-brake relay, and EBS hard stop.

Merging this after those would **roll the pin backwards** to v1.0.0 and silently undo that stack, so it is safer closed than left open. Nothing is lost — everything v1.0.0 contained is an ancestor of the current pin.

Reopen if I've misread the intent.

